PR #68, opened 2021-11-15
Fix up routing resolution behavior

andrewstucki:
This PR adds support for the `PrefixPath` routing that was, unfortunately overlooked during initial development. Additionally, there was a bug with the way we were tracking `HTTPRoute` rules that have multiple rule entries, where the services specified as `backendRefs` across the entire set of `rules` get combined into the final rule specified, and config entries were created solely for this rule.

The fixes:

1. Add a map entry for `PrefixPath` in our `HTTPRoute` --> internal representation code
2. Make a copy of the route rule as we're resolving services -- the iterating loop was taking an address of a value returned via a `range` operation, which was making all of the map keys the same.

How I've tested this PR:

Added unit tests and verified manually.

Checklist:
- [x] Tests added

### DIFF
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -1,5 +1,3 @@
-//+build e2e
-
 package server
 
 import (
@@ -255,7 +253,6 @@ func TestMeshService(t *testing.T) {
 			routeOneName := envconf.RandomName("route", 16)
 			routeTwoName := envconf.RandomName("route", 16)
 			routeThreeName := envconf.RandomName("route", 16)
-			routeFourName := envconf.RandomName("route", 16)
 
 			gatewayNamespace := gateway.Namespace(namespace)
 			resources := cfg.Client().Resources(namespace)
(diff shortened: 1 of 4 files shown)


andrewstucki:
Anyone running across this, current workaround is to use path-based `RegularExpression` rules separated onto multiple routes rather than a single route with multiple matches an `PathPrefix`

internal/commands/server/k8s_e2e_test.go
```diff
@@ -392,6 +393,22 @@ func TestMeshService(t *testing.T) {
 								},
 							},
 						}},
+					}, {
+						BackendRefs: []gateway.HTTPBackendRef{{
```
andrewstucki, line 397: This exercises the code path that was failing previously by making an `HTTPRoute` with multiple rules such that the `for` loop with the  previously questionable range + address usage gets iterated over twice. The collapsed routing table should be the same as before the change to the testing code.

mikemorris (approved):
Confirmed the e2e test catches this case now, fix LGTM.

PR #69, opened 2021-11-17
Initial TCPRoute Implementation

andrewstucki:
This PR introduces the Gateway API's `TCPRoute`s -- in the process of implementation I also found a bug in how we were setting binding/resolution errors in our route statuses and went ahead and fixed that.

There are two implementation-specific constraints for `TCPRoute`s that I added. Due to the way Consul `IngressGateway` config entries only support single services when they leverage the TCP protocol:

1. `TCPRoute`s may only have a single backend service, which means
  a. Only a single routing rule can be defined on a `TCPRoute`
  b. The single routing rule can only have a single entry in `backendRefs`
2. The spec doesn't specify what to do with multiple `TCPRoute`s that reference the same Gateway listener, because of the need for only a single backend, we mark the listener status with `Conflicted` condition of `RouteConflict`, which essentially keeps the listener from being bound on the gateway until its resolved.

Checklist:
- [x] Tests added

Still adding more tests and cleaning stuff up, so going to mark this as a draft for the time being.

### DIFF
--- a/internal/k8s/reconciler/route.go
+++ b/internal/k8s/reconciler/route.go
@@ -188,10 +188,24 @@ func (r *K8sRoute) OnBindFailed(err error, gateway store.Gateway) {
 					status.Accepted.ListenerNamespacePolicy = err
 				case BindErrorTypeRouteKind:
 					status.Accepted.InvalidRouteKind = err
+				case BindErrorTypeRouteInvalid:
+					status.Accepted.BindError = err
 				}
-				return
+			} else {
+				status.Accepted.BindError = err
 			}
-			status.Accepted.BindError = err
+			// set resolution errors - we can do this here because
+			// a route with resolution errors will always fail to bind
+			errorType, err := r.resolutionErrors.Flatten()
+			switch errorType {
+			case service.GenericResolutionErrorType:
+				status.ResolvedRefs.Errors = err
+			case service.ConsulServiceResolutionErrorType:
+				status.ResolvedRefs.ConsulServiceNotFound = err
+			case service.K8sServiceResolutionErrorType:
+				status.ResolvedRefs.ServiceNotFound = err
+			}
+
 			r.parentStatuses[id] = status
 		}
 	}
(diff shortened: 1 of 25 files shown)


internal/k8s/reconciler/route.go
```diff
@@ -312,6 +327,14 @@ func (r *K8sRoute) Resolve(listener store.Listener) *core.ResolvedRoute {
 			"consul-api-gateway/k8s/HTTPRoute.Name":      r.GetName(),
 			"consul-api-gateway/k8s/HTTPRoute.Namespace": r.GetNamespace(),
 		}, route, r)
+	case *gw.TCPRoute:
+		return convertTCPRoute(namespace, prefix, map[string]string{
+			"external-source":                           "consul-api-gateway",
```
nickethier, line 332: nit: we could probably DRY up meta generation

internal/k8s/reconciler/route.go
```diff
@@ -354,6 +377,36 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
...
+	case *gw.TCPRoute:
+		if len(route.Spec.Rules) != 1 {
+			err := service.NewResolutionError("a single tcp rule is required")
+			r.resolutionErrors.Add(err)
+			return nil
+		}
+
+		rule := route.Spec.Rules[0]
+
+		if len(rule.BackendRefs) != 1 {
```
nickethier, line 389: This is a constraint imposed by Consul right? We should comment that at a minimum. Idk if we've already started a list of spec deviations but this should probably be documented somewhere more visible (eventually).

internal/k8s/controllers/http_route_controller.go
```diff
@@ -28,7 +28,7 @@ type HTTPRouteReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
 func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := r.Log.With("route", req.NamespacedName)
+	logger := r.Log.With("http-route", req.NamespacedName)
```
nickethier, line 31: Maybe we should just make this `route-name` since the type (http vs tcp) is already in the logger name

nickethier (approved):
A couple nits/thoughts as I was reading through. Looks good!

internal/k8s/reconciler/route.go
andrewstucki: Yep, this is a Consul constraint as you can only specify a single service in an `ingress_gateway` config entry for TCP services and service splitters don't allow for TCP -- so we basically can support 1 TCPRoute/gateway, with 1 routing rule and 1 backend ref -- it's unfortunate, but our current constraint. 👍 on the documentation comment.

PR #70, opened 2021-12-16
Enabling Security-Scan for CRT

claire-labry:
Changes proposed in this PR:

What this PR does is it enables security scanning (owned by the Security team) for this repository. This is a part of the Common Release Tooling (CRT) effort `which enables a consistent workflow for security scanning that can be used across products and services. It provides a single abstraction point to allow us integrate our own scanning capabilities, as well as leverage external tools and APIs from vendors, open source, or free-to-use projects.` Learn more about HashiCorp's security-scanner [here](https://github.com/hashicorp/security-scanner/).

If anyone needs any further clarification or if nothing makes sense, feel free to comment or reach out to me! ❤️

How I've tested this PR:

I ran the tests in crt-workflows-common and the passing tests are as follows: 

- [security-scan-binaries](https://github.com/hashicorp/crt-workflows-common/actions/runs/1588311739)
- [security-scan-containers](https://github.com/hashicorp/crt-workflows-common/actions/runs/1588318571)
- [verify](https://github.com/hashicorp/crt-workflows-common/actions/runs/1588330432) -- this is the very last workflow that is run in workflows common and verifies that everything looks good. 

You can test this PR by creating a new branch by setting your branch name in `release_branches` in `ci.hcl` and changing the `push` event in `build.yml` to your branch name -- this will cause the build workflow to run, once succeeded, the orchestrator will kick off the workflows in [crt-workflows-common](https://github.com/hashicorp/crt-workflows-common/actions).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,8 +3,7 @@ name: build
 on:
   push:
     branches:    
-      # - 'release/**'
-      - 'enable-security-scan'
+      - 'release/**'
 
 env:
   CGO_ENABLED: 0
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,8 @@ name: build
 on:
   push:
     branches:    
-      - 'release/**'
+      # - 'release/**'
+      - 'enable-security-scan'
 
 env:
   CGO_ENABLED: 0
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -3,12 +3,12 @@ schema = "1"
 project "consul-api-gateway" {
   team = "consul-api-gateway"
   slack {
-    notification_channel = "C01RWVBQ6GJ"
+    notification_channel = "C01A3A54G0L"
   }
   github {
     organization = "hashicorp"
     repository = "consul-api-gateway"
-    release_branches = ["release/0.1.x"]
+    release_branches = ["enable-security-scan"]
   }
 }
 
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -98,4 +98,4 @@ jobs:
           target: default
           arch: ${{matrix.arch}}
           tags: |
-            docker.io/hashicorp/${{env.repo}}:${{env.version}}
\ No newline at end of file
+            docker.io/hashicorp/${{env.repo}}:${{env.version}}
```
claire-labry, line 101: I don't know what happened here, but I can confirm nothing changed here. 🤷‍♀️

sarahethompson: I think it's just because the new line at the end of the file has been removed. Adding it back should remove this from the diff!

sarahethompson (approved):
LGTM!

.github/workflows/build.yml
claire-labry: I double checked the editor and there was an newline and if you edit this file you will see that there's a newline 🙈 

I could be way off base or just simply not seeing something? :bushes:

andrewstucki: looks like the original yaml file was missing the newline, most folks editors normalize by adding a newline onto the end of most files anyway, so I'm fine with the appended line break, it'll likely happen sooner or later anyway.

andrewstucki (approved):
this looks fine to me if the security-scan action-specific configuration looks right on the releng side of things

PR #71, opened 2021-12-20
add support for configurable TLS params

mikemorris:
Depends on https://github.com/hashicorp/consul/pull/11576

Changes proposed in this PR:
- Move `Certificates` from top-level config on `ResolvedListener` and `K8sListener` structs into `TLSParams` struct.
- Adds specific tests for `TLSParams` parsing in `k8s/reconciler`
- Updates `ConsulSyncAdapter` to apply TLS params from `ResolvedGatreway` and adds a new `ConsulSyncAdapter_Sync` test

How I've tested this PR:
- [x] Added k8s/reconciler tests
- [x] Added `adapters/consul` sync test
- [x] Added e2e test
- [x] Manually tested e2e workflow following the new Learn tutorial

Checklist:
- [x] Tests added
- [ ] ~~CHANGELOG entry added~~ we will adopt a consistent changelog maintenance strategy before GA, likely implementing https://github.com/hashicorp/go-changelog


internal/k8s/reconciler/listener_test.go
```diff
@@ -179,6 +179,125 @@ func TestListenerValidate(t *testing.T) {
...
+		},
+	}, K8sListenerConfig{
+		Logger: hclog.NewNullLogger(),
+		Client: client,
+	})
+	client.EXPECT().GetSecret(gomock.Any(), gomock.Any()).Return(&core.Secret{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "secret",
+		},
+	}, nil)
```
mikemorris, line 296: Is there a good way to template all this boilerplate and allow just passing a custom map in `TLS.Options`?

internal/k8s/reconciler/listener.go
```diff
@@ -119,12 +140,86 @@ func (l *K8sListener) validateTLS(ctx context.Context) error {
...
+	"TLS_AUTO": {}, // Remove if Envoy ever sets TLS 1.3 as default minimum
+	"TLSv1_0":  {},
+	"TLSv1_1":  {},
+	"TLSv1_2":  {},
+}
+
+// FIXME: import this from Consul API
+var supportedTlsCipherSuites = map[string]struct{}{
+	"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256": {},
+}
```
mikemorris, line 221: Would it be appropriate to import these lookup tables from the Consul API? I'd hate the have to maintain them independently, although having them in some shared Envoy/xDS config dependency will probably make more sense in the future than accessing through Consul core API.

internal/k8s/reconciler/listener.go
```diff
@@ -119,12 +140,109 @@ func (l *K8sListener) validateTLS(ctx context.Context) error {
...
+			}
+
+			// set cipher suites on listener TLSParams
+			if l.tlsParams != nil {
+				l.tlsParams.cipherSuites = tlsCipherSuites
+			} else {
+				l.tlsParams = &tlsParams{
+					cipherSuites: tlsCipherSuites,
+				}
+			}
```
mikemorris, line 217: Checking this for every possible option is a bit tedious, should `K8sListener.tlsParams` maybe not be a pointer?

internal/core/resolved.go
```diff
@@ -21,13 +21,20 @@ type ResolvedRoute interface {
...
 	Certificates []string
-	Routes       []ResolvedRoute
+}
+
+type ResolvedListener struct {
+	Name      string
+	Hostname  string
+	Port      int
+	Protocol  string
+	TLSParams *TLSParams
```
mikemorris, line 36: Should this actually not be a pointer?

internal/k8s/reconciler/listener.go
mikemorris: Changed this to not be a pointer when switching to `core.TLSParams`

internal/k8s/reconciler/listener.go
andrewstucki: I suppose it all depends on how these are exposed in Consul. I like the idea of importing from there to keep things a bit cleaner, at least so long as we don't diverge from what the core codebase supports -- but are these exposed in the Consul subpackage?

internal/core/resolved.go
andrewstucki: So, the main time that pointers actually really matter in a structure field is if you have to distinguish between "not supplied" and the zero-value of the type. In this case does `TLSParams` zero-value reflect what we want to use when no value is specified?

internal/k8s/reconciler/listener.go
```diff
@@ -119,12 +125,85 @@ func (l *K8sListener) validateTLS(ctx context.Context) error {
...
+var tlsVersionsWithConfigurableCipherSuites = map[string]struct{}{
+	"TLS_AUTO": {}, // Remove if Envoy ever sets TLS 1.3 as default minimum
+	"TLSv1_0":  {},
+	"TLSv1_1":  {},
+	"TLSv1_2":  {},
+}
+
+// FIXME: import this from Consul API
+var supportedTlsCipherSuites = map[string]struct{}{
+	"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256": {},
```
Jeff-Apple, line 204: If this setting the default cipher suites enabled on the Listener? If so, the listener needs a much larger list than this for the default cipher suites.

mikemorris: Yes, this was just a placeholder for testing - I was hoping to have a single source of truth for this list exposed through the Consul API but that doesn't seem like it will be possible. In the future it would be nice to have this list as part of a common xDS library used by both Consul core and Consul API Gateway.

internal/k8s/reconciler/listener.go
mikemorris: I really wanted to expose these, but the Consul codebase doesn't allow sharing anything between `consul` and `consul/api` (even things like types or struct defs) from either package to the other, which seems to make this not possible. In the future it would be nice to have this list as part of a common xDS library used by both Consul core and Consul API Gateway.

internal/adapters/consul/sync_test.go
```diff
@@ -143,3 +145,50 @@ func TestHTTPRouteDiscoveryChain(t *testing.T) {
...
+	// consulSrv.WaitForServiceIntentions(t)
+
+	// FIXME: Config entry not found for "ingress-gateway" / "name1"
+	entry, _, err := consul.ConfigEntries().Get(api.IngressGateway, "name1", nil)
+	require.NoError(t, err)
+	ingress, ok := entry.(*api.IngressGatewayConfigEntry)
+	require.True(t, ok)
+	require.NotNil(t, ingress)
+	fmt.Printf("%#v\n", ingress)
+	require.Equal(t, "TLSv1_2", ingress.TLS.TLSMinVersion)
```
mikemorris, line 193: Is there a better way to wait for the "sync" operation to complete than using require.Eventually here? I noticed that consul/api has some WaitFor functions available, not sure if anything like that would be useful here.

internal/adapters/consul/sync.go
```diff
@@ -182,12 +182,26 @@ func discoveryChain(gateway core.ResolvedGateway) (*api.IngressGatewayConfigEntr
 
 		if len(services) > 0 {
 			tls := &api.GatewayTLSConfig{}
-			if len(listener.Certificates) > 0 {
+
+			if listener.TLS.MinVersion != "" {
+				tls.TLSMinVersion = listener.TLS.MinVersion
+			}
```
mikemorris, line 188:
```suggestion
			if listener.TLS.MinVersion != "" {
				tls.TLSMinVersion = listener.TLS.MinVersion
			} else {
			        // set secure default instead of Envoy's TLS 1.0 default
			        tls.TLSMinVersion = "TLSv1_2"
			}
```

internal/adapters/consul/sync.go
```diff
@@ -182,12 +182,26 @@ func discoveryChain(gateway core.ResolvedGateway) (*api.IngressGatewayConfigEntr
...
+
+			if listener.TLS.MaxVersion != "" {
+				tls.TLSMaxVersion = listener.TLS.MaxVersion
+			}
+
+			if len(listener.TLS.CipherSuites) > 0 {
+				tls.CipherSuites = listener.TLS.CipherSuites
+			}
```
mikemorris, line 196:
```suggestion
			if len(listener.TLS.CipherSuites) > 0 {
				tls.CipherSuites = listener.TLS.CipherSuites
			} else {
			        defaultCipherSuites := make([]string, 0, len(supportedTlsCipherSuites))
    			        for k := range supportedTlsCipherSuites {
			                defaultCipherSuites = append(keys, k)
			        }
			        tls.CipherSuites = defaultCipherSuites
			}
```

internal/store/interfaces.go
```diff
@@ -45,7 +45,10 @@ type ListenerConfig struct {
 	Hostname string
 	Port     int
 	Protocol string
-	TLS      bool
+
+	// TODO: is this field a sufficient reason to use TLSParams instead of TLS as the
+	// field name on Listener structs to avoid confusion?
+	TLS bool
```
mikemorris, line 51: Highlighting this question.

internal/k8s/reconciler/listener.go
```diff
@@ -123,12 +129,118 @@ func (l *K8sListener) validateTLS(ctx context.Context) error {
...
+// TODO: move this somewhere where it can be imported as default config in
+// internal/adapters/consul when CipherSuites is undefined
+var defaultTlsCipherSuites = []string{
+	"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
+	"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
+	"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+	"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+	"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+	"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+}
```
mikemorris, line 216: Is there a common location this could live to be imported by both `k8s/reconciler` and `adapters/consul`?

andrewstucki: Generally it depends on how tied it is to either k8s or consul. If it's highly specific to `consul` because it maps to the config entry values, maybe the `consul` subpackage, otherwise things that are shared logic but not tied to any particular integration, but more just shared internally in our `consul-api-gateway` world, I've been putting under `common`.

internal/store/interfaces.go
andrewstucki: Yeah, probably, it'd make the term overloading a bit less heavy.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -702,24 +704,43 @@ func TestTCPMeshService(t *testing.T) {
...
+						// 		CertificateRefs: []*gateway.SecretObjectReference{{
+						// 			Name:      "consul-server-cert",
+						// 			Namespace: &gatewayNamespace,
+						// 		}},
+						// 		Options: map[gateway.AnnotationKey]gateway.AnnotationValue{
+						// 			"api-gateway.consul.hashicorp.com/tls_min_version":   "TLSv1_1",
+						// 			"api-gateway.consul.hashicorp.com/tls_cipher_suites": "TLS_RSA_WITH_AES_128_CBC_SHA",
+						// 		},
+						// 	},
+						// },
```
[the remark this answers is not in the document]
andrewstucki: So, unfortunately TCPRoutes, due to the way they are implemented at the Consul level can only route to a single backend. I think I have a comment somewhere about this in the code, but the explanation for sure can be found in the implementation PR: https://github.com/hashicorp/consul-api-gateway/pull/69

That said, I'm fine with either writing another test or just hijacking the options of the already deployed route in this test. If you do go with another test though, you'll want to allocate another bound port and punch yet another hole in the test `kind` configuration to be able to reach the service from the testing code. Let me know how you want to proceed and I can walk you through it.

internal/store/interfaces.go
mikemorris: Would changing this field to `TLSEnabled` or something be a viable alternative?

andrewstucki: I'd be fine with that, so long as it makes sense, I'm even ok collapsing it into an `Enabled` boolean within a larger `TLS` configuration structure -- the corresponding code that uses the field currently would need to be updated though. Currently these are all just used as in-memory representations of our configuration that map to the underlying consul config entries and k8s manifests, so it should be fine to make the change, and frankly I'd prefer to do that now rather than in the future if we have say, a persistent data store for some of this stuff.

mikemorris:
It looks like the e2e tests are currently failing with the following error, and I'm not quite sure the most direct way to debug this as I _think_ I updated to the latest commit at https://github.com/hashicorp/consul/commits/release/1.11.x/api correctly - is there maybe something that needs to be regenerated I missed?
```
   |       "reason": "SyncError",
  |       "message": "error adding ingress config entry: 1 error occurred:\n\t* Unexpected response code: 400 (Bad request: Request decoding failed: 2 errors occurred:\n\t* invalid config key \"Listeners[0].TLS.TLSMinVersion\"\n\t* invalid config key \"Listeners[0].TLS.CipherSuites\")\n\n"
  ```

File: internal/commands/server/k8s_e2e_test.go [hunk 12]
```diff
@@ -702,52 +709,76 @@ func TestTCPMeshService(t *testing.T) {
...
+			// Force insecure cipher suite excluded from Consul API Gateway default
+			// cipher suites, but supported by Envoy defaults
+			//
+			// FIXME: why is this test not failing? attempting to connect with only
+			// this cipher suite should be rejected by the Consul API Gateway listener
+			// default config
+			checkTCPTLSRoute(t, listenerOnePort, &tls.Config{
+				InsecureSkipVerify: true,
+				CipherSuites:       []uint16{tls.TLS_RSA_WITH_AES_128_CBC_SHA},
+			}, serviceOne.Name, "service not routable in allotted time")
```
mikemorris, line 762: Oof, so it looks like I may have been encoding this incorrectly in the Consul > Envoy xDS config generation, I was using a comma-delimited string of cipher suites when it looks like the correct syntax appears to be `"[SUITE_ONE|SUITE_TWO]"`, thus causing my invalid config to be ignored in favor of Envoy's defaults 😭

andrewstucki: the beauty of what e2e tests catch in this world of multi-service distributed deployments 😬

mikemorris: Hmm, maybe that comment from 2018 was using v2 syntax, it does look like a [repeated string](https://www.envoyproxy.io/docs/envoy/v1.20.1/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-tlsparameters) type in v3

mikemorris: Oops, this config was being ignored because the TLS client and server negotiated TLS 1.3, which doesn't allow manually configuring cipher suites.

internal/store/interfaces.go
mikemorris: Think I was able to pretty cleanly refactor this into the TLSParams struct as an `Enabled bool` field.

internal/adapters/consul/sync_test.go
```diff
@@ -202,6 +202,7 @@ func TestConsulSyncAdapter_Sync(t *testing.T) {
 		require.True(t, ok)
 		require.NotNil(t, ingress)
 
-		return ingress.Listeners[0].TLS.TLSMinVersion == "TLSv1_2"
+		return ingress.Listeners[0].TLS.TLSMinVersion == "TLSv1_2" &&
+			reflect.DeepEqual(ingress.Listeners[0].TLS.CipherSuites, common.DefaultTLSCipherSuites())
```
andrewstucki, line 206: do we know if Consul maintains the ordering here? `reflect.DeepEqual` returns `false` if the order of an array changes IIRC

internal/commands/server/k8s_e2e_test.go
```diff
@@ -751,15 +751,14 @@ func TestTCPMeshService(t *testing.T) {
...
-			// default config
+			// cipher suites, but supported by Envoy defaults, limit max version to
+			// TLS 1.2 to ensure cipher suite config is applicable. The error returned
+			// is a bit odd, but is expected for this failure case.
 			checkTCPTLSRoute(t, listenerOnePort, &tls.Config{
 				InsecureSkipVerify: true,
+				MaxVersion:         tls.VersionTLS12,
 				CipherSuites:       []uint16{tls.TLS_RSA_WITH_AES_128_CBC_SHA},
-			}, serviceOne.Name, "service not routable in allotted time")
+			}, "remote error: tls: unknown certificate authority", "service not routable in allotted time")
```
andrewstucki, line 761: can you change the expectation message here? it seems like your expectation is an error with this message, correct? probably should say something like `expected error not received` or something like that, may even want to do something like wrap the error in the helper function so we actually know what happened.

internal/adapters/consul/sync_test.go
mikemorris: The sync adapter assigns these directly to the config entry at https://github.com/hashicorp/consul-api-gateway/pull/71/files#diff-9defe7ecca2a3bf51ee560c795c29ba0d8c95dfef3ef8c864aeab9e85a5ec97fR199 and AFAIK Consul should preserve the ordering of values in an array (as opposed to something like map key iteration order which won't be guaranteed).

Envoy appears to rely on [BoringSSL's cipher suite negotiation](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/ssl_cipher.cc) which ensures predefined ordering to prefer more secure cipher suites so preserving the ordering here isn't _essential_ as it may be for other proxy config, but I don't believe it should change.

internal/commands/server/k8s_e2e_test.go
mikemorris: Oops, so adding `t.Log` error logging clearly showed that I should've been expecting a different error message `"remote error: tls: handshake failure"`, which actually makes a lot more sense.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -934,13 +1011,14 @@ func checkTCPTLSRoute(t *testing.T, port int, expected string, message string) {
...
-		tlsConn := tls.Client(conn, &tls.Config{
-			InsecureSkipVerify: true,
-		})
+		tlsConn := tls.Client(conn, config)
 		data, err := io.ReadAll(tlsConn)
+
 		if err != nil {
-			return false
+			t.Log(err)
+			return strings.HasPrefix(err.Error(), expected)
```
mikemorris, line 1019: Should the other `check*` functions be updated similarly to also be able to check for expected error states?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -934,13 +1011,14 @@ func checkTCPTLSRoute(t *testing.T, port int, expected string, message string) {
...
 		}
```
mikemorris, line 1020: Should the other `check*` functions be updated similarly to also be able to check for expected error states?

andrewstucki: Originally the `check*` functions didn't take into account errors, only passing if no error was there, i.e. has `require.NoError` implicit in it. I'd just keep it the way it is for now unless you are actually touching them in the updated tests. We can always do a subsequent refactor down the line.

internal/k8s/reconciler/gateway.go
```diff
@@ -68,7 +68,8 @@ func NewK8sGateway(gateway *gw.Gateway, config K8sGatewayConfig) *K8sGateway {
 func (g *K8sGateway) certificates() []string {
 	certificates := []string{}
 	for _, listener := range g.listeners {
-		certificates = append(certificates, listener.Certificates()...)
+		listenerTlsParams := listener.Config().TLS
+		certificates = append(certificates, listenerTlsParams.Certificates...)
```
mikemorris, line 72:
```suggestion
		certificates = append(certificates, listener.Config().TLS.Certificates...)
```

mikemorris:
**tl;dr** Confirmed e2e functionality by walking through the upcoming Learn guide with an image built from this branch deployed.

---

Testing in a browser showed the following default negotiated version and cipher suite selection:
```
{
	"Connection:": {
		"Protocol version:": "TLSv1.3",
		"Cipher suite:": "TLS_AES_128_GCM_SHA256",
		"Key Exchange Group:": "x25519",
		"Signature Scheme:": "ECDSA-P256-SHA256"
	}
}
```

After applying the following settings to the API Gateway YAML config, it successfully limited negotiation to an older TLS version and specific cipher suite:
```
    tls:
      options:
        "api-gateway.consul.hashicorp.com/tls_max_version": "TLSv1_2"
        "api-gateway.consul.hashicorp.com/tls_cipher_suites": "TLS_RSA_WITH_AES_128_CBC_SHA"
```
```
{
	"Connection:": {
		"Protocol version:": "TLSv1.2",
		"Cipher suite:": "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA",
		"Key Exchange Group:": "x25519",
		"Signature Scheme:": "ECDSA-P256-SHA256"
	}
}
```

Setting a TLS max version of 1.1 on the gateway (and lowering the min version to 1.0) displayed the following error in the browser when attempting to connect.
```
Error code: SSL_ERROR_PROTOCOL_VERSION_ALERT
This website might not support the TLS 1.2 protocol, which is the minimum version supported by Firefox. Enabling TLS 1.0 and TLS 1.1 might allow this connection to succeed.

TLS 1.0 and TLS 1.1 will be permanently disabled in a future release.
```

Setting a single cipher suite unsupported by the browser (I tested with `TLS_RSA_WITH_AES_128_CBC_SHA`) displayed the following error in Firefox:
```
Error code: SSL_ERROR_NO_CYPHER_OVERLAP
```

Attempting to lower the max version below the default min, or specify cipher suites for a min version of TLS 1.3 both set error statuses on the gateway as expected.
```
      Message:               error adding ingress config entry: 1 error occurred:
                           * Unexpected response code: 500 (rpc error making call: configuring max version TLSv1_1 less than the configured min version TLSv1_2 is invalid)
```
```
      Message:               configuring TLS cipher suites is only supported for TLS 1.2 and earlier
```

internal/common/tls.go
```diff
@@ -0,0 +1,45 @@
...
+	"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+	"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+	"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+}
+
+func DefaultTLSCipherSuites() []string {
+	return defaultTLSCipherSuites
+}
+
+// NOTE: the following cipher suites are currently supported by Envoy but insecure and
```
andrewstucki, line 16: Note that we might have some issues here since we allow specifying which version of envoy you want to use for a gateway as part of the `GatewayClassConfig` CRD. Not that we can exactly handle version checking at the controller level since you could even specify a custom envoy container/build, but we may want to document in the actual docs that this option is potentially particularly sensitive to the envoy version that is being deployed for the gateway.

andrewstucki (approved):
This looks pretty good to me. One thing mainly just to note in our public-facing docs, but not anything that sticks out to me that requires change before merge.

PR #72, opened 2022-01-10
use configured controller CA in gateway deployments

nickethier:
This PR moves deployment and service building to a separate package and support injecting a Consul CACert into gateway instances at runtime.


internal/k8s/builder/gateway.go
```diff
@@ -0,0 +1,330 @@
...
+// the exec container.
+const gwContainerCommandTpl = `
+{{- if .ConsulCAFile}}
+export CONSUL_CACERT={{ .ConsulCAFile }}
+cat <<EOF >{{ .ConsulCAFile }}
+{{ .ConsulCAData }}
+EOF
+{{- end}}
+
+/bootstrap/consul-api-gateway exec -log-json \
```
andrewstucki, line 317: To make sure that the signal handling actually works:

```suggestion
exec /bootstrap/consul-api-gateway exec -log-json \
```

otherwise you'll have to do some signal forwarding trickery since all of this is being invoked as a child process currently with `sh -c` -- the extra `exec` will avoid the extra child process and make sure any termination signals go directly to our binary.

andrewstucki (approved):

PR #73, opened 2022-01-12
Fix Consul 1.11 certificate write blocking issues

andrewstucki:
The description of the issue is in the comment on the workaround/fix -- we basically force a blocking query timeout prior to kicking off the cert watching goroutines to make sure we successfully retrieve the leaf cert prior to blocking until rotation. Communicated the issue upstream so that we can get a fix in core as well.

How I've tested this PR:

Deployed a gateway talking to Consul 1.11 -- restarted the gateway a few times and made sure it always retrieved the certificates.

Workaround for https://github.com/hashicorp/consul/issues/12048

### DIFF
--- a/internal/consul/certmanager.go
+++ b/internal/consul/certmanager.go
@@ -122,8 +122,9 @@ type CertManager struct {
 	rootWatch *watch.Plan
 	leafWatch *watch.Plan
 
-	// this can be overwritten to check retry logic in testing
-	writeCerts certWriter
+	// these can be overwritten to modify retry logic in testing
+	writeCerts     certWriter
+	skipExtraFetch bool
 }
 
 // NewCertManager creates a new CertManager instance.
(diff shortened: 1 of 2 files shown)


internal/consul/certmanager.go
```diff
@@ -246,6 +250,25 @@ func (c *CertManager) Manage(ctx context.Context) error {
...
+	// cause all subsequent non-blocking queries to unexpectedly block. The problem
+	// is that this means that, on restart, the query for a leaf certificate with
+	// the given service id will never return until the previous leaf certificate
+	// expires/is rotated. Adding a wait here causes the API to return once the timeout has
+	// been hit -- allowing us to short-circuit the buggy blocking. The subsequent
+	// goroutines can then be leveraged to pick up any certificate rotations.
+	if !c.skipExtraFetch {
+		leafCert, _, err := c.consul.Agent().ConnectCALeaf(c.service, &api.QueryOptions{
+			WaitTime: 1 * time.Second,
+		})
```
FFMMM, line 266: Wondering if churning / repeated calls here would be possible.

I don't have a lot of context on the project here btw, just a thought.

andrewstucki: Just to make sure I'm understanding -- is it a question about whether or not we'd repeatedly call this code or a suggestion to wrap this in a retry loop?

If the former, this code path should get called exactly once at the beginning of the process leveraging it (in this case our controller and gateway instances which use the CA endpoints to establish a trusted SDS connection), the goroutines that get spawned after this leverage `api/watch` to make repeated blocking query-based updates when the connect certs rotate.

Since the bug is really only a problem when we first boot up due to hanging when `api/watch` first kicks off its non-blocking request (which become blocking requests due to the bug), we essentially just jam this here to do a one-off retrieval before the blockng kicks in.

mikemorris: My understanding from the discussion was "if the RPC call takes longer than the timeout (even if it doesn’t block) it could return an emtpy result" so it would be important to retry here if needed?

andrewstucki: So, in this particular case it I believe it should still be fine -- if the result is empty due to a timeout, the background watch should pick up the certificate since it has no timeout on it (hence why the non-blocking call became blocking previously). I know there are some oddities around this working the way it does, but it is a workaround v. ideally just using the watch code.

PR #74, opened 2022-01-13
Add service account management

andrewstucki:
Changes proposed in this PR:

Allow for an option on the `GatewayClassConfig` object to create a service account per-gateway deployment (useful for scoped consul permissions via auth methods)

How I've tested this PR:

```diff
diff --git a/dev/config/k8s/consul-api-gateway.yaml b/dev/config/k8s/consul-api-gateway.yaml
index d952bd1..87599a3 100644
--- a/dev/config/k8s/consul-api-gateway.yaml
+++ b/dev/config/k8s/consul-api-gateway.yaml
@@ -10,6 +10,8 @@ spec:
   consul:
     scheme: https
     caSecret: consul-ca-cert
+    authentication:
+      managed: true
     ports:
       http: 8501
       grpc: 8502
```

Run:

```bash
➜  consul-api-gateway git:(managed-service-account) ✗ kubectl apply -f dev/config/k8s/consul-api-gateway.yaml
...
➜  consul-api-gateway git:(managed-service-account) ✗ kubectl get pod -o json $(kubectl get pods | grep gateway | head -n 1 | cut -d' ' -f1) | grep \"serviceAccount\"
        "serviceAccount": "test-gateway",
➜  consul-api-gateway git:(managed-service-account) ✗ kubectl get serviceaccount test-gateway
NAME           SECRETS   AGE
test-gateway   1         4m30s
➜  consul-api-gateway git:(managed-service-account) ✗ kubectl delete -f dev/config/k8s/consul-api-gateway.yaml
...
➜  consul-api-gateway git:(managed-service-account) ✗ kubectl get serviceaccount test-gateway
Error from server (NotFound): serviceaccounts "test-gateway" not found
```

### DIFF
--- a/pkg/apis/v1alpha1/types.go
+++ b/pkg/apis/v1alpha1/types.go
@@ -117,6 +117,8 @@ type CopyAnnotationsSpec struct {
 }
 
 type AuthSpec struct {
+	// Whether deployments should be run with "managed" service accounts created by the gateway controller.
+	Managed bool `json:"managed,omitempty"`
 	// The Consul auth method used for initial authentication by consul-api-gateway.
 	Method string `json:"method,omitempty"`
 	// The Kubernetes service account to authenticate as.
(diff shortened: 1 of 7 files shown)


pkg/apis/v1alpha1/types.go
```diff
@@ -117,6 +117,8 @@ type CopyAnnotationsSpec struct {
 }
 
 type AuthSpec struct {
+	// Whether deployments should be run with "managed" service accounts created by the gateway controller.
+	Managed bool `json:"managed,omitempty"`
```
nickethier, line 121: Is the `Account` field still relevant if `Managed` is true? Should we validate that both are not set?

nickethier (approved):

pkg/apis/v1alpha1/types.go
andrewstucki: I can submit a follow-up PR that documents a little bit better which takes priority, but no, both aren't still relevant. At the same time, I'm not so sure I'd want to consider the config invalid if both are specified, but I'm sure we can hash that out subsequently.

PR #86, opened 2022-02-02
Allow API Gateway controller to update k8s Deployments + Services

nathancoleman:
The k8s client abstraction in consul-api-gateway [expects](https://github.com/hashicorp/consul-api-gateway/blob/c82e707fb30bebf1a88980568e0f67e318f65496/internal/k8s/gatewayclient/gatewayclient.go#L67-L68) to be able to create **or update** k8s `Deployments` and `Services`; however, the `ClusterRole` as defined today does not allow for updates to these resources. This prevents the Consul API Gateway controller from successfully installing a new gateway to the cluster.

Changes proposed in this PR:
- Modify `ClusterRole` to allow controller to update k8s `Deployment`
- Modify `ClusterRole` to allow controller to update k8s `Service`

How I've tested this PR:
Applied changes to deployment on GKE

Follow the [Learn tutorial](https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway) for Consul API Gateway; however, the Consul Helm chart install needs to use the branch in this PR.

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > Entries should use present tense (e.g. Add support for...)

### DIFF
--- a/internal/k8s/controllers/gateway_controller.go
+++ b/internal/k8s/controllers/gateway_controller.go
@@ -36,8 +36,8 @@ type GatewayReconciler struct {
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch
-//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=list;get;create;watch
-//+kubebuilder:rbac:groups=core,resources=services,verbs=list;get;create;watch
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=list;get;create;update;watch
+//+kubebuilder:rbac:groups=core,resources=services,verbs=list;get;create;update;watch
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=list;get;create;watch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
 
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -29,6 +29,7 @@ rules:
   - create
   - get
   - list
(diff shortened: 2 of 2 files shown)


internal/k8s/controllers/gateway_controller.go
```diff
@@ -36,8 +36,8 @@ type GatewayReconciler struct {
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch
-//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=list;get;create;watch
-//+kubebuilder:rbac:groups=core,resources=services,verbs=list;get;create;watch
```
andrewstucki, line 40: Good catch, with our current mechanism for e2e testing these sorts of RBAC issues are hard to test as we use the same credentials for the controller as the superuser permissions created by the `kind create` command. Do you mind creating a follow-up issue for tracking purposes of modifying our e2e test suite to use a reduced set of permissions via the generated RBAC policy?

andrewstucki (approved):
Clarification that this does allow for _new_ gateways to be deployed, but subsequent modifications to the gateway that affect the deployment/service configurations such as an annotation change on the gateway or listener port changes while the gateway has an associated service deployment should fail without these permission updates. Aside from me commenting on the scope of this, LGTM.

nathancoleman:
> Clarification that this does allow for new gateways to be deployed ...

I was thinking the same; however, my `Gateway` resource would never actually reach a `ready=true` state when running `$ kubectl apply --filename api-gw/consul-api-gateway.yaml` due to this issue.

nathancoleman:
Myself, @andrewstucki, and @mikemorris are spending some more time together looking into this since it seems GKE is needing to update a `Service` when we wouldn't expect it to; however, the nature of the change is applicable regardless of what we find.

PR #91, opened 2022-02-08
Enable namespace mapping for gateway service registration and future mesh service lookup

andrewstucki:
This PR introduces namespace mapping similar to the options allowed in our `consul-k8s` helm chart:

https://www.consul.io/docs/k8s/helm#v-synccatalog-consulnamespaces

In order to simplify our proposed `MeshService` CRD to rely on the same namespace mapping security posture as `consul-k8s` we essentially need to support the same k8s --> consul namespace mappings that they do. This allows k8s namespace-based RBAC policies to control access to particular consul-namespace services.

By doing this for the future `MeshService` CRD it essentially means we just need a consul service name and the consul namespace will be constructed based off of the k8s namespace that the CRD resource is created in.

Prior to this all gateways were just created in the default namespace, this also fixes that.


david-yu:
Thanks for doing this @andrewstucki. Those same docs apply to connectInject as well, which I what I believe is what you are targeting as opposed to catalogSync for API Gateway: https://www.consul.io/docs/k8s/helm#v-connectinject-consulnamespaces

nickethier:
I'm :+1: for the general approach here. It wasn't clear to me if you need to opt into namespace mapping. Would be good to see some unit tests around making sure we don't pass Consul a namespace (through gateway exec) accidentally unless explicitly configured to do so.

mikemorris:
@david-yu Are those docs missing a mention of how k8s services are mapped into `consulDestinationNamespace` when `mirroringK8S: false` - isn't there some sort of k8s-namespace-prefixing of service names that happens, or is the configured elsewhere?

david-yu:
@mikemorris Prefixes are only used if `mirroringK8S` is set to `true`. https://www.consul.io/docs/k8s/helm#v-connectinject-consulnamespaces-consuldestinationnamespace says that `consuldestinationnamespace` is ignored if mirroring is set to true.

internal/k8s/controller.go
```diff
@@ -43,6 +43,19 @@ func init() {
...
+	MirrorKubernetesNamespaces      bool
+	MirrorKubernetesNamespacePrefix string
+}
+
+func (c ConsulNamespaceConfig) Namespace(namespace string) string {
+	if c.MirrorKubernetesNamespaces {
+		return c.MirrorKubernetesNamespacePrefix + namespace
+	}
+	return c.ConsulDestinationNamespace
+}
```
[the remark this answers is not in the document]
andrewstucki: So, the main problem I see with build tags, unless they modify some code that's generating namespaces, is that if we say, have some separate distinct tests, we're basically going to want to duplicate all of our testing code in the enterprise tagged build. Personally I'd rather have a few high-level e2e tests that leverage a good amount of our feature set than treat the e2e tests like cheap unit tests.

What I think would be preferable is doing something like generating a namespace of `""` for our OSS Consul deployments, a random namespace for our enterprise deployments, and then adding some code into our e2e tests that takes this namespace into account when verifying the registration on the Consul side of things. This would at least test that our namespace interactions with Consul are working. Our mapping function leveraging prefixes v. static namespaces v. k8s namespaces can just be unit tested from my perspective.

internal/commands/exec/exec.go
```diff
@@ -86,6 +86,14 @@ func RunExec(config ExecConfig) (ret int) {
 			return 1
 		}
 		config.Logger.Trace("consul login complete")
+	} else {
+		consulConfig := &config.ConsulConfig
+		consulConfig.Namespace = config.GatewayConfig.Namespace
```
andrewstucki, line 91: I can add a comment here if y'all feel it necessary, but basically this is a bit of a hack to get our leaf cert setup working. Internally we're leveraging https://github.com/hashicorp/consul/tree/main/api/watch for grabbing and rotating leaf/root certs from the connect CA. It, unfortunately, has no concept of namespaces.

Since we're trying to match up the existence of a gateway based on the certificate it's pulled from connect, we need the leaf cert to be pulled using a Namespace parameter, this gets us that.

internal/commands/exec/exec.go
```diff
@@ -205,7 +214,9 @@ func login(config ExecConfig) (*api.Client, string, error) {
 
 	// Now update the client so that it will read the ACL token we just fetched.
 	config.ConsulConfig.Token = token
-	newClient, err := api.NewClient(&config.ConsulConfig)
+	consulConfig := &config.ConsulConfig
+	consulConfig.Namespace = config.GatewayConfig.Namespace
```
andrewstucki, line 218: Same as above.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -32,8 +32,9 @@ import (
 )
 
 var (
-	testenv   env.Environment
-	hostRoute string
+	testenv      env.Environment
+	hostRoute    string
+	timeoutCheck = 1 * time.Minute
```
andrewstucki, line 37: bumped this up since, for some reason the enterprise version of consul, at least locally, seems to take longer registering services and pushing down xDS, which causes the timeouts in our `require.Eventually` expectations to get hit

internal/envoy/testdata/basic.golden.json
```diff
@@ -10,7 +10,10 @@
   },
   "node": {
     "cluster": "6f164d52-7846-45d1-97b1-fc984795572b",
-    "id": "6f164d52-7846-45d1-97b1-fc984795572b"
+    "id": "6f164d52-7846-45d1-97b1-fc984795572b",
+    "metadata": {
+      "namespace": "default"
```
andrewstucki, line 15: see https://github.com/hashicorp/consul/blob/c17aa3720fbbc4033bd41a5986fc41af8e7b142e/command/connect/envoy/bootstrap_tpl.go#L142-L145

internal/testing/e2e/docker.go
```diff
@@ -17,6 +19,10 @@ type dockerImageContext struct{}
 
 var dockerImageContextKey = dockerImageContext{}
 
+const (
+	envvarExtraDockerImages = envvarPrefix + "DOCKER_IMAGES"
```
andrewstucki, line 23: this lets me use a custom build of consul enterprise locally by building out-of-band and then specifying:

```
E2E_APIGW_CONSUL_IMAGE=consul-dev:ent
E2E_APIGW_DOCKER_IMAGES=consul-dev:ent
```
in the test run

internal/commands/server/k8s_e2e_test.go
```diff
@@ -245,7 +250,8 @@ func TestHTTPMeshService(t *testing.T) {
 			require.NoError(t, err)
 			serviceTwo, err := e2e.DeployHTTPMeshService(ctx, cfg)
 			require.NoError(t, err)
-			serviceThree, err := e2e.DeployHTTPMeshService(ctx, cfg)
+			// register this service in a different consul namespace
+			serviceThree, err := e2e.DeployHTTPMeshService(ctx, cfg, e2e.ConsulNamespace(ctx))
```
andrewstucki, line 254: this makes it so we're actually leveraging routing between multiple namespaces from a single gateway in our tests

internal/testing/e2e/consul.go
```diff
@@ -564,10 +601,29 @@ func gatewayConsulRole(namespace, policyID string) *api.ACLRole {
...
+	if IsEnterprise() {
+		return &api.ACLPolicy{
+			Name: "consul-api-gateway",
+			Rules: `
+	namespace_prefix "" {
+	acl = "write"
+	policy = "write"
+	service_prefix "" { policy = "write" }
+	session_prefix "" { policy = "write" }
+	node_prefix "" { policy = "read" }
```
nathancoleman, line 613: Can we indent these to make it easier to read, or will that break things?

internal/testing/e2e/consul.go
```diff
@@ -530,6 +540,33 @@ func CreateConsulAuthMethod(namespace string) env.Func {
...
+		consulEnvironment := ctx.Value(consulTestContextKey)
+		if consulEnvironment == nil {
+			return ctx, nil
+		}
+		env := consulEnvironment.(*consulTestEnvironment)
+		_, _, err := env.consulClient.Namespaces().Create(&api.Namespace{
+			Name: namespace,
+		}, &api.WriteOptions{
+			Token: env.token,
+		})
```
nathancoleman, line 561: I trip all over this formatting, but `gofmt` prefers it this way

internal/testing/e2e/docker.go
```diff
@@ -17,6 +19,10 @@ type dockerImageContext struct{}
 
 var dockerImageContextKey = dockerImageContext{}
 
+const (
```
nathancoleman, line 22: For my own knowledge, do we use parens for single-line declarations going forward? I've seen some inconsistency but wasn't sure what our team convention is.

internal/k8s/service/resolver.go
```diff
@@ -218,6 +222,9 @@ func (r *backendResolver) Resolve(ctx context.Context, ref gw.BackendObjectRefer
 			return nil, NewK8sResolutionError("service port must not be empty")
 		}
 		return r.consulServiceForK8SService(ctx, namespacedName)
+	case group == apigwv1alpha1.GroupVersion.Group && kind == "MeshService":
+		r.findConsulMeshService(namespacedName)
+		fallthrough
```
nathancoleman, line 227: What's the reasoning for `fallthrough` to a `NewResolutionError` here?

nathancoleman (changes requested):

internal/commands/exec/exec.go
andrewstucki: Yeah, good call, I can just change this to `config.ConsulConfig.Namespace = config.GatewayConfig.Namespace` and keep the old `NewClient` code as it was previously.

internal/k8s/service/resolver.go
```diff
@@ -218,6 +222,9 @@ func (r *backendResolver) Resolve(ctx context.Context, ref gw.BackendObjectRefer
 			return nil, NewK8sResolutionError("service port must not be empty")
 		}
 		return r.consulServiceForK8SService(ctx, namespacedName)
+	case group == apigwv1alpha1.GroupVersion.Group && kind == "MeshService":
```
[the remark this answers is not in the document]
andrewstucki: Yep, we do have a constant for it, but it's in the subsequent PR, this is just a placeholder for the stubbing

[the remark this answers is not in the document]
nathancoleman: I noticed that once I got into reviewing #92 but forgot to come back here and address my previous comment 😆

internal/testing/e2e/service.go
```diff
@@ -267,17 +274,21 @@ type bootstrapArgs struct {
...
-func DeployTCPMeshService(ctx context.Context, cfg *envconf.Config) (*core.Service, error) {
-	return deployMeshService(ctx, cfg, "tcp", tcpBootstrapTemplate)
+func DeployTCPMeshService(ctx context.Context, cfg *envconf.Config, consulNamespace ...string) (*core.Service, error) {
+	return deployMeshService(ctx, cfg, "tcp", tcpBootstrapTemplate, consulNamespace...)
 }
 
-func deployMeshService(ctx context.Context, cfg *envconf.Config, protocol string, template *template.Template) (*core.Service, error) {
+func deployMeshService(ctx context.Context, cfg *envconf.Config, protocol string, template *template.Template, consulNamespaces ...string) (*core.Service, error) {
+	consulNamespace := ""
+	if len(consulNamespaces) != 0 {
```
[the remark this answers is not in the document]
andrewstucki: This is just to check whether or not any parameters were passed in. It's a bit gross, in terms of the implementation code, but makes it so we can just have a variadic function for deploying the services, if you don't specify a namespace, it'll deploy it to the default namespace, if you pass in a namespace parameter at the end, it'll grab that and use it.

internal/testing/e2e/docker.go
andrewstucki: not sure we really have a consistent convention yet, generally I prefer `const` blocks to make it easier to add constants to in the future

internal/testing/e2e/docker.go
mikemorris: This functionality doesn't seem to be working for me in local testing, not quite sure for what reason yet - whether there's actually an issue here or if I'm perhaps skipping a step somewhere. @nathancoleman any luck validating this?

internal/testing/e2e/kind.go
```diff
@@ -227,19 +227,31 @@ func LoadKindDockerImage(clusterName string) env.Func {
...
+			}
 		}
 
 		return ctx, nil
 	}
 }
+
+func loadImage(ctx context.Context, clusterName, image string) error {
+	var stdout, stderr bytes.Buffer
+	timeoutContext, cancel := context.WithTimeout(ctx, 20*time.Second)
```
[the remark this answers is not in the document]
mikemorris: [As someone without much Go familiarity] would reassigning `ctx` introduce any issues with tracing or cancellation, where the passed-in `context.Context` is discarded/clean up earlier than expected?

internal/testing/e2e/docker.go
mikemorris: I've only used parens for multiple-delcaration, I don't _think_ I've seen parens used for singular much at all.

internal/k8s/builder/gateway.go
```diff
@@ -81,11 +80,12 @@ func filterAnnotations(annotations map[string]string, allowed []string) map[stri
...
-	gwConfig     *v1alpha1.GatewayClassConfig
-	sdsHost      string
-	sdsPort      int
-	consulCAData string
+	gateway          *gw.Gateway
+	gatewayNamespace string
+	gwConfig         *v1alpha1.GatewayClassConfig
+	sdsHost          string
+	sdsPort          int
+	consulCAData     string
```
mikemorris, line 88:
```suggestion
	gateway                *gw.Gateway
	gwConfig               *v1alpha1.GatewayClassConfig
	sdsHost                string
	sdsPort                int
	consulCAData           string
	consulGatewayNamespace string
```
For places if/where we're referencing Consul namespaces instead of k8s namespaces, I think it would be really helpful to clarify that in variable/struct field naming. (I'm not sure if this is actually one of those places though.)

internal/k8s/service/resolver.go
```diff
@@ -290,26 +297,54 @@ func (r *backendResolver) findGlobalCatalogService(service *corev1.Service) (*Re
...
-		if err != nil {
-			r.logger.Trace("error retrieving node services", "error", err, "node", node.Node)
+
+	var consulNamespaces []*api.Namespace
+	// the default namespace
+	namespaces := []string{""}
+	consulNamespaces, _, err = r.consul.Namespaces().List(nil)
+	if err != nil {
+		if !strings.Contains(err.Error(), "Unexpected response code: 404") {
+			r.logger.Trace("error retrieving namespaces", "error", err)
```
mikemorris, line 307: Could this return other errors like a 5xx if ACLs are wrong? (The string matching instead of checking a status code feels 😬 but IIRC that's an upstream issue with `consul/api`)

internal/testing/e2e/consul.go
```diff
@@ -530,6 +540,33 @@ func CreateConsulAuthMethod(namespace string) env.Func {
 	}
 }
 
+func IsEnterprise() bool {
+	return strings.HasSuffix(consulImage, "ent")
```
mikemorris, line 544: This feels like a kinda sketchy check, but fine for now. I'm curious if there's a better to do something like query supported features.

internal/testing/e2e/kind.go
```diff
@@ -227,19 +227,31 @@ func LoadKindDockerImage(clusterName string) env.Func {
...
-		cmd := exec.CommandContext(timeoutContext, "kind", "load", "docker-image", image, image, "--name", clusterName)
-		cmd.Stderr = &stderr
-		cmd.Stdout = &stdout
-		if err := cmd.Run(); err != nil {
-			return nil, errors.New(stderr.String())
+		for _, image := range ExtraDockerImages() {
+			log.Printf("Loading additional docker image:%s into kind cluster", image)
+			if err := loadImage(ctx, clusterName, image); err != nil {
+				return nil, err
+			}
```
mikemorris, line 238: Running `go test ./... -tags e2e -run TestHTTPMeshService` locally for me produces the following output, which looks like (at best) the same image is attempting to be loaded twice, but the e2e test also seems to consistently time out 😞 
```
2022/02/17 17:19:49 Loading docker image into kind cluster
2022/02/17 17:20:09 Loading additional docker image:consul-dev:ent into kind cluster
2022/02/17 17:20:29 Image: "consul-dev:ent" with ID "sha256:8492c3d4c6dcd4404bc0a20f232e5f7bac73f14c4f724cf1c7db9aabb03184c8" not yet present on node "consul-api-gateway-test-016573-control-plane", loading...
Image: "consul-dev:ent" with ID "sha256:8492c3d4c6dcd4404bc0a20f232e5f7bac73f14c4f724cf1c7db9aabb03184c8" not yet present on node "consul-api-gateway-test-016573-control-plane", loading...
```
I'm guessing with this code, specifying _both_ env vars isn't actually necessary?
```
E2E_APIGW_CONSUL_IMAGE=consul-dev:ent
E2E_APIGW_DOCKER_IMAGES=consul-dev:ent
```

mikemorris: Welp, it looks like maybe the `-run` flag was obscuring the actual issue 🤦 

I modified `build-support/docker/Consul-Dev.dockerfile` to add a test license 
```
COPY test.lic test.lic
RUN chmod +r test.lic
ENV CONSUL_LICENSE_PATH=./test.lic
```
Which _works fine_ when running `docker run -t --rm consul-dev:ent` but I'm guessing maybe it isn't being copied/set in the `consul-api-gateway` image we're building. Would be fine to add extra tooling to support the explicit license file in a folllowup PR.

mikemorris (approved):
Only real comments are around naming - figured out the kind image load issue, we can add support for loading a Consul Enterprise license file in a separate PR if needed.

internal/testing/e2e/kind.go
andrewstucki: yeah, so I believe Nick did some work to set up the license mounting in our test setu to work with a `CONSUL_LICENSE` env variable that gets leveraged in CI, see:
 https://github.com/hashicorp/consul-api-gateway/blob/dc7562ac7226e582e162c3b0a548163e5715283a/internal/testing/e2e/consul.go#L30

I'm thinking you can probably just `export CONSUL_LICENSE="$(cat ./test.lic)"`.

RE the environment variables, you _will_ need both -- the first is the Consul image to run in the container deployed in the testing `kind` cluster (which should be your locally built image), and the `E2E_APIGW_DOCKER_IMAGES` is a little shim that does a `kind load` of said image prior to deploying Consul.

internal/testing/e2e/consul.go
andrewstucki: Yeah, it's definitely a hack, and not proud of it. Main alternative I can think of is doing something like just adding a `consul_ent.go` with a build tag that overrides `IsEnterprise` or something of the sort and add a `go test -tag ent` to pick it up. Alternatively I'd love if Consul had the equivalent of a `version` api or something of the sort that gave you back whether or not you had an enterprise deployment or not, but I don't think there is, only some other gross hacks like testing for the existence of APIs

internal/k8s/service/resolver.go
andrewstucki: in which case that's completely fine, the 404 is mainly due to the fact that OSS consul will give a 404 for the entirety of that endpoint -- it's what I mentioned as a dirty hack to tell whether you're using OSS or enterprise -- basically if we don't have that API endpoint we should still loop once over the empty "namespace" since we're on OSS

internal/commands/server/command.go
```diff
@@ -71,6 +76,14 @@ func (c *Command) init() {
 	c.flagSet.IntVar(&c.flagSDSServerPort, "sds-server-port", defaultSDSServerPort, "SDS Server Port.")
 	c.flagSet.IntVar(&c.flagMetricsPort, "metrics-port", 0, "Metrics port, if not set, metrics are not enabled.")
 	c.flagSet.IntVar(&c.flagPprofPort, "pprof-port", 0, "Go pprof port, if not set, profiling is not enabled.")
+
+	{
+		// Consul namespaces
+		c.flagSet.StringVar(&c.flagConsulDestinationNamespace, "consul-destination-namespace", "", "Consul namespace to register gateway services.")
+		c.flagSet.BoolVar(&c.flagMirrorK8SNamespaces, "mirroring-k8s", false, "Register Consul gateway services based on Kubernetes namespace.")
+		c.flagSet.StringVar(&c.flagMirrorK8SNamespacePrefix, "mirroring-k8s-prefix", "", "Namespace prefix for Consul services when mirroring Kubernetes namespaces.")
```
nathancoleman, line 84: Assuming we don't expect these to conflict with any future flags by dropping the `-namespaces` suffix?

nathancoleman (approved):

internal/testing/e2e/kind.go
andrewstucki: Line 230 is loading up a docker image of the current codebase so that it can be deployed, it's not controlled by any env variables, and always needs to happen. This other bit is to load up any additional images that your environment may need.

internal/commands/server/command.go
andrewstucki: Yeah, I don't imagine there are going to be any other flags regarding "mirroring", but the values I just dasherized the option names from helm, which are [mirroringK8S](https://www.consul.io/docs/k8s/helm#v-synccatalog-consulnamespaces-mirroringk8s) and [mirroringK8SPrefix](https://www.consul.io/docs/k8s/helm#v-synccatalog-consulnamespaces-mirroringk8sprefix).

internal/k8s/service/resolver.go
mikemorris: Ooh, okay - that could probably use a comment explaining what's going on there.

internal/testing/e2e/consul.go
andrewstucki: Yeah, not that I don't have a conflict of interest, but am kind of partial to something like what Elasticsearch does, exposing version info at a top-level endpoint. Assuming you have auth turned on this could be protected to avoid folks banner grabbing:

```bash
➜  ~ curl http://localhost:9200
{
  "name" : "f9b9188b4715",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "PU5ca1ZFRBGuXmYmCPlE9A",
  "version" : {
    "number" : "8.0.0",
    "build_flavor" : "default",
    "build_type" : "docker",
    "build_hash" : "1b6a7ece17463df5ff54a3e1302d825889aa1161",
    "build_date" : "2022-02-03T16:47:57.507843096Z",
    "build_snapshot" : false,
    "lucene_version" : "9.0.0",
    "minimum_wire_compatibility_version" : "7.17.0",
    "minimum_index_compatibility_version" : "7.0.0"
  },
  "tagline" : "You Know, for Search"
}
```

david-yu:
@andrewstucki Awesome work! Out of curiosity I saw this PR, is this still outstanding? https://github.com/hashicorp/consul-k8s/pull/1024

andrewstucki:
@david-yu Yes, that's just a quick addition that was pending this merge, mainly because the options didn't make sense until this was finalized (and sure enough, the flag names changed from the initial PR). So just marked that as ready for review.

PR #90, opened 2022-02-07
Include service namespace and name in resolution errors

nathancoleman:
Fixes #87

Changes proposed in this PR:
- Include the namespace and name of the service for which resolution failed

How I've tested this PR:

How I expect reviewers to test this PR:

Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > HashiCorp engineers only, community PRs should not add a changelog entry.
  > Entries should use present tense (e.g. Add support for...)

PR #92, opened 2022-02-08
Add support for non-k8s services

andrewstucki:
This is the spiritual successor to https://github.com/hashicorp/consul-api-gateway/pull/24 and https://github.com/hashicorp/consul-api-gateway/pull/22.

Based on https://github.com/hashicorp/consul-api-gateway/pull/91 we can have proper support for a `MeshService` CRD that simply specifies a consul service to route to on the mesh.

As this is a stacked PR and the other is still in a "Draft" state until we get namespace-based testing enabled for our e2e tests, opening this as a draft as well.

# Manual testing

If you want to manually test this the easiest way is probably without the additional namespace mapping code. Since the new CRD allows you to specify a mesh-enabled service directly, just go about the standard connect-inject workflow, but:

1) create a `MeshService` object that points to the Consul service with the name of the service as mapped in Consul
2) instead of referencing a k8s service in your route `backendRefs`, override the group/kind specification to be `group: "api-gateway.consul.hashicorp.com"` and `kind: "MeshService"` and specify the `MeshService` object you created in step 1


internal/commands/server/k8s_e2e_test.go
```diff
@@ -665,6 +680,7 @@ func TestTCPMeshService(t *testing.T) {
 		}).
 		Assess("tls routing", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			serviceOne, err := e2e.DeployTCPMeshService(ctx, cfg)
+			require.NoError(t, err)
```
nathancoleman, line 683: Good catch!

internal/k8s/service/resolver.go
```diff
@@ -222,9 +222,8 @@ func (r *backendResolver) Resolve(ctx context.Context, ref gw.BackendObjectRefer
 			return nil, NewK8sResolutionError("service port must not be empty")
 		}
 		return r.consulServiceForK8SService(ctx, namespacedName)
-	case group == apigwv1alpha1.GroupVersion.Group && kind == "MeshService":
-		r.findConsulMeshService(namespacedName)
-		fallthrough
+	case group == apigwv1alpha1.GroupVersion.Group && kind == apigwv1alpha1.MeshServiceKind:
```
nathancoleman, line 225: :heart:

internal/k8s/service/resolver.go
```diff
@@ -321,9 +320,83 @@ func (r *backendResolver) findGlobalCatalogService(service *corev1.Service) (*Re
 	return nil, nil
 }
 
-// this will resolve a service based on our future CRD
-func (r *backendResolver) findConsulMeshService(name types.NamespacedName) {
-	mapped := r.mapper(name.Namespace)
+func (r *backendResolver) consulServiceForMeshService(ctx context.Context, namespacedName types.NamespacedName) (*ResolvedReference, error) {
+	var err error
+	var resolved *ResolvedReference
```
nathancoleman, line 325: Could use named return values here, but I haven't noticed whether that's a thing we do/don't do in this repo

internal/k8s/service/resolver.go
```diff
@@ -321,9 +320,83 @@ func (r *backendResolver) findGlobalCatalogService(service *corev1.Service) (*Re
...
+
+	service, err := r.client.GetMeshService(ctx, namespacedName)
+	if err != nil {
+		r.logger.Trace("error retrieving mesh service", "error", err, "name", namespacedName.Name, "namespace", namespacedName.Namespace)
+		return nil, err
+	}
+	if service == nil {
+		return nil,
+			NewConsulResolutionError(fmt.Sprintf(
+				"kubernetes mesh service object (%q, %q) not found", namespacedName.Name, namespacedName.Namespace,
```
nathancoleman, line 335: We wound up just using the namespaced name directly in https://github.com/hashicorp/consul-api-gateway/pull/90 and below inside the retry. I'd like to see us settle on a convention for this

nathancoleman (commented):
The Kubernetes parts look good here; however, I'm still getting up to speed on the Consul landscape so would like to make sure @mikemorris reviews as well

nathancoleman:
How difficult do you think it'd be to add test coverage here?

andrewstucki:
it wouldn't be too hard now with the additional testing work handled in #91, the main reason I didn't have it previously is because some of the implementation was blocked on the namespace code, but I can add an end-to-end test in the morning for it.

internal/k8s/service/resolver.go
andrewstucki: yeah, that should be fine, the string output isn't much different, think the `NamespacedName`'s Stringer implementation outputs the equivalent of replacing: `(%q, %q)` with `{%s, %s}`

internal/k8s/service/resolver.go
andrewstucki: I'm not really a fan of named returns unless you're using at least 3 return values, but maybe that's just me, I remember their being some org-wide style guidelines that we may want to follow here, but I'm assuming either way there's not totally a hard or fast rule on it.

nathancoleman: It's not something I feel strongly about, just a thought I had as I read through

mikemorris (approved):
Still working through manual testing, but reviewed and implementation LGTM

internal/k8s/service/resolver.go
```diff
@@ -342,9 +341,83 @@ func (r *backendResolver) findGlobalCatalogService(service *corev1.Service) (*Re
...
+	consulName := service.Spec.Name
+	services, _, err := r.consul.Catalog().Service(consulName, "", &api.QueryOptions{
+		Namespace: consulNamespace,
+	})
+	if err != nil {
+		r.logger.Trace("error resolving consul service reference", "error", err)
+		return nil, err
+	}
+	if len(services) == 0 {
+		return nil, NewConsulResolutionError(fmt.Sprintf("consul service (%s, %s) not found", consulNamespace, consulName))
```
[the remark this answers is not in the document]
andrewstucki: Since we're pulling the namespace from the mapping code and the name from the resolved `MeshService` parameter, I'd prefer to avoid an additional struct initialization just for formatting sake -- that said, might be worth doing a follow-up PR going through this file and the rest of the codebase to settle on some standardized logging and error messaging formats.

config/crd/bases/api-gateway.consul.hashicorp.com_meshservices.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: MeshService holds a reference to an externally managed Consul
+          Service Mesh service.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
nathancoleman, line 28:
```suggestion
              internal value and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```

nathancoleman (approved):

config/crd/bases/api-gateway.consul.hashicorp.com_meshservices.yaml
andrewstucki: that's an auto-generated comment, so you'll need to take it up with the kubebuilder folks upstream 😁

PR #93, opened 2022-02-08
Add check for clean code generation in CI

andrewstucki:
This PR adds a step in our CI to make sure that all generated code (apart from test fixtures, which should fail tests) are up-to-date.


nathancoleman:
@andrewstucki @mikemorris Should this include #89 or should this be merged first?

nickethier (approved):

.github/workflows/ci.yml
```diff
@@ -48,6 +48,15 @@ jobs:
         product: consul
         version: ${{ matrix.consul-version }}
 
+    - name: Clean Generate
+      run: |
+        echo "Checking that code generation is up-to-date"
+        go install github.com/golang/mock/mockgen@v1.6.0
+        go generate ./...
```
mikemorris, line 55: Should this be captured in a make task instead of installing the specifc mockgen version only in CI?

mikemorris (approved):
Aside from one question on `mockgen` LGTM

.github/workflows/ci.yml
andrewstucki: we could extract it into a `make` task like the others if we so desire, not opposed to the idea -- in fact I'd almost personally be inclined to just make a single `make generate` target that does all of these tasks rather than have to run each of the targets individually, personally think that it makes for better dev ergonomics

PR #94, opened 2022-02-08
run e2e against enterprise consul binary

nickethier:
This PR adds a matrix build to the e2e ci step and plumbs the necessary information into the e2e environment to boot an enterprise consul cluster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: ["main", "release/**"]
 env:
-  GO_VERSION: '1.16'
+  GO_VERSION: '1.17'
   CONSUL_LICENSE: ${{ secrets.CONSUL_LICENSE }}
 jobs:
   lint:
@@ -66,10 +66,13 @@ jobs:
         path: ${{ env.TEST_RESULTS_DIR }}
 
   e2e:
-    name: e2e tests
+    name: e2e tests (consul-image=${{ matrix.consul-image }})
+    strategy:
+      matrix:
+        consul-image: ['hashicorp/consul:1.11.2', 'hashicorp/consul-enterprise:1.11.2-ent']
     runs-on: ubuntu-latest
     env:
-      TEST_RESULTS_DIR: /tmp/test-results/e2e
+      TEST_RESULTS_DIR: /tmp/test-results/e2e@${{ matrix.consul-image }}
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


internal/testing/e2e/consul.go
```diff
@@ -287,6 +292,13 @@ func consulDeployment(namespace string, httpsPort, grpcPort int) *apps.Deploymen
 	labels := map[string]string{
 		"deployment": "consul-test-server",
 	}
+	var env []core.EnvVar
+	if consulEntLicense != "" {
```
andrewstucki, line 296: Just wondering, since the exercise of namespace code is going to require figuring out if we're running an enterprise v. non-enterprise version of Consul, should we have something to distinguish between the two? My thoughts are either an appended `ent` build tag on the test run with some `ent`-specific testing code or an environment variable we set to determine "this is an enterprise run". I'm assuming that this license env variable is set regardless of the consul image being run?

.github/workflows/ci.yml
```diff
@@ -66,10 +66,13 @@ jobs:
         path: ${{ env.TEST_RESULTS_DIR }}
 
   e2e:
-    name: e2e tests
+    name: e2e tests (consul-image=${{ matrix.consul-image }})
+    strategy:
+      matrix:
+        consul-image: ['hashicorp/consul:1.11.2', 'hashicorp/consul-enterprise:1.11.2-ent']
```
nathancoleman, line 72: TIL 🤩 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategy

nathancoleman (approved):

internal/testing/e2e/consul.go
nickethier: That is correct about the license being present in CI during oss and ent runs.

I was going to play with writing ent tests in a file with a separate build tag and working out a way to include the tag during the ent jobs.

internal/testing/e2e/consul.go
```diff
@@ -24,8 +25,10 @@ import (
 )
 
 const (
-	consulImage          = "hashicorp/consul:1.11.2"
-	configTemplateString = `
+	defaultConsulImage            = "hashicorp/consul:1.11.2"
```
mikemorris, line 28: Would it be possible to pull this default up a few layers to somewhere more obvious, or maybe even remove it and require it to be set explicitly? Any thoughts on using something like https://github.com/joho/godotenv for default testing env?

nathancoleman: godotenv reminded me of [rakyll/globalconf](https://github.com/rakyll/globalconf) which I've used in almost every microservice I've built over the last few years. The beauty is that you can provide config values in `.ini` files, flags to your run command, or in the environment. Each value maps to a stdlib flag in your application code.

PR #100, opened 2022-02-15
Prevent stale GatewayClassConfig

### LINKED ISSUE #88: GatewayClassConfig can become stale
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

It appears as though our dirty-checking for `GatewayClassConfig`s has a bug and our `GatewayClassConfig` values can become stale.

### Reproduction Steps

- Deploy the gateway controller via `helm` with a managed gateway class config.
- Update the helm deployment with a slightly different configuration for the managed class config (i.e. change the `serviceType` value)
- Deploy a gateway
- You should see the old class config being used and serialized onto the Gateway annotations, meaning we're holding a stale reference to the configuration

Note: restarting the controller and re-creating the gateway uses the latest class config.

nathancoleman:
Fixes #88 

Changes proposed in this PR:
- When a `GatewayClassConfig` changes, evict all `GatewayClasses` that reference it from the cache

How I've tested this PR:
See instructions in #88 

Do the following on both main and this branch, observing that the config annotation on the `Gateway` is correct on this branch:
```shell
$ ./dev/run

# In a separate terminal
$ kubectl apply -f ./dev/config/k8s/

# Observe config in Gateway annotation
$ kubectl get gateway test-gateway -o yaml

# Add serviceType to GatewayClassConfig
$ kubectl edit gatewayclassconfig test-gateway-class-config

# Delete the gateway and recreate it, expecting that it will pick up the new serviceType in the config annotation
$ kubectl delete gateway test-gateway
$ kubectl apply -f ./dev/config/k8s/

# Observe config in Gateway annotation
# On main, the added serviceType will not be included at this stage.
# On this branch, the serviceType will be included.
$ kubectl get gateway test-gateway -o yaml

# Quit the ./dev/run process, switch branches, and then
$ ./dev/run -r

# Start back at the first command
```

Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 
  > Entries should use present tense (e.g. Add support for...)

### DIFF
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -6,6 +6,7 @@ package server
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -15,6 +16,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
@@ -65,7 +67,7 @@ func TestGatewayBasic(t *testing.T) {
 			resources := cfg.Client().Resources(namespace)
 
 			gatewayName := envconf.RandomName("gw", 16)
-			_, gc := createGatewayClass(ctx, t, cfg)
+			gcc, gc := createGatewayClass(ctx, t, cfg)
 
 			require.Eventually(t, func() bool {
 				created := &gateway.GatewayClass{}
@@ -126,6 +128,8 @@ func TestGatewayBasic(t *testing.T) {
 				}
 				return false
 			}, 30*time.Second, 1*time.Second, "no gateway found in the allotted time")
+
+			checkGatewayConfigAnnotation(t, created, gcc)
 
 			// check for the service being registered
 			client := e2e.ConsulClient(ctx)
@@ -963,6 +967,19 @@ func createTCPRoute(ctx context.Context, t *testing.T, resources *resources.Reso
 
 	err := resources.Create(ctx, route)
 	require.NoError(t, err)
+}
+
+// checkGatewayConfigAnnotation verifies that the GatewayClassConfig was
+// correctly serialized into the expected annotation on the Gateway.
+func checkGatewayConfigAnnotation(t *testing.T, g *gateway.Gateway, gcc *apigwv1alpha1.GatewayClassConfig) {
+	t.Helper()
+
+	expectedCfg, err := json.Marshal(gcc.Spec)
+	require.NoError(t, err)
+
+	actualCfg, ok := g.Annotations[`api-gateway.consul.hashicorp.com/config`]
+	assert.True(t, ok)
+	assert.Equal(t, string(expectedCfg), actualCfg)
 }
 
 func checkRoute(t *testing.T, port int, path, expected string, headers map[string]string, message string) {
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -73,7 +73,7 @@ func TestGatewayWithClassConfigChange(t *testing.T) {
 
 			// Create a Gateway and wait for it to be ready
 			firstGatewayName := envconf.RandomName("gw", 16)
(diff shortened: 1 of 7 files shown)


nathancoleman:
@mikemorris what should I be doing with the `CHANGELOG` here? It looks outdated, but I know you were looking into an automated process.

internal/k8s/controllers/gateway_class_config_controller.go
```diff
@@ -62,6 +64,16 @@ func (r *GatewayClassConfigReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}
 
 	// we're creating or updating
+
+	// evict any class that's referencing an updated config from cache
+	if using, err := r.Client.GatewayClassesUsingConfig(ctx, gcc); err == nil && len(using.Items) > 0 {
+		for _, gc := range using.Items {
+			if err := r.Manager.DeleteGatewayClass(ctx, gc.Name); err != nil {
+				logger.Warn("error evicting cached gateway class referencing config", "error", err)
```
nathancoleman, line 72: @mikemorris @andrewstucki this is best effort right now. Should we keep it that way, or should we be erroring out if we potentially have a stale `GatewayClassConfig` in the cache?

nathancoleman (commented):
The e2e test runs seem to be intermittently timing out. I can't imagine that being caused by this change, but I'm not sure if it's something anyone else has seen since we merged those last week. @mikemorris @andrewstucki ?

mikemorris:
> @mikemorris what should I be doing with the `CHANGELOG` here? It looks outdated, but I know you were looking into an automated process.

Started implementation in https://github.com/hashicorp/consul-api-gateway/pull/101 but shouldn't block this, we can always add an entry afterwards.

internal/k8s/controllers/gateway_class_config_controller.go
andrewstucki: think we should be fine right now

andrewstucki (commented):
I would love part of an e2e test that essentially leverages what we were seeing before -- think you can do something like create a `GatewayClassConfig`, create `GatewayClass` that leverages it, update the `GatewayClassConfig`, deploy a `Gateway` and verify that the serialized config on the deployed gateway matches the updated config. Think it shouldn't be too bad just to plumb into something existing.

nathancoleman:
@andrewstucki added some e2e coverage for this fix. My additions are passing; however, one of the existing tests is not. Seems possibly related to what you're actively working on - could you take a look?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -880,6 +873,41 @@ func conditionInSync(conditions []meta.Condition) bool {
...
+	gw := &gateway.Gateway{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      gatewayName,
+			Namespace: namespace,
+		},
+		Spec: gateway.GatewaySpec{
+			GatewayClassName: gateway.ObjectName(gc.Name),
+			Listeners: []gateway.Listener{{
+				Name:     "https",
+				Port:     gateway.PortNumber(443),
```
andrewstucki, line 893: @nathancoleman So this is why your tests are failing. When you extracted this helper you set all of the listeners to bind to port `443` -- the problem is that to actually route through the listener that's deployed in `kind` from the localhost you need to do all sorts of port-forwarding in the `kind` cluster itself. If you want to keep this extracted helper, then you'll need to parameterize it to take a port and then pass in `e2e.HTTPPort(ctx)` to the Gateway that you actually want to check the routing through, which I believe is currently on line 302.

andrewstucki (changes requested):
You'll need to fix up the `createGateway` helper method you created to fix up the tests, but almost there!

internal/commands/server/k8s_e2e_test.go
nathancoleman: Good catch!

andrewstucki (approved):

PR #106, opened 2022-02-20
Fix kubernetes-sigs/gateway-api/config version v0.4.0 to v0.5.0-dev

clock21am:
Changes proposed in this PR:
Fix kubernetes-sigs/gateway-api/config version v0.4.0 to v0.5.0-dev as version v0.4.0 is no longer available so updated to current stable version

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -2,6 +2,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.4.1
+- github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.5.0-dev
 - bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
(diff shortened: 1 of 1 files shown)


config/crd/kustomization.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.4.1
+- github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.5.0-dev
```
nathancoleman, line 5: @clock21am This `v0.5.0-dev` ref doesn't seem to exist. Is this something you expect to exist soon?

nathancoleman (changes requested):

nathancoleman:
I appreciate you taking a look at this over the weekend @clock21am!

The `v0.4.1` ref that we're using today can be accessed [here](https://github.com/kubernetes-sigs/gateway-api/tree/v0.4.1/config/crd), and there is no `v0.5.0-dev` ref available today. Since we're already using the latest available ref, I'm going to close this PR. Please comment and let us know if you believe we've missed anything.

PR #107, opened 2022-02-23
envoy: bump default image to v1.20-latest

mikemorris:
I don't particularly like how this default is buried a few directories deep, but reorganizing that feels out of scope for now.

**Changes proposed in this PR:**
- Bumps the default Envoy image used to `envoyproxy/envoy:v1.20-latest`

**How I've tested this PR:** 

**How I expect reviewers to test this PR:**

**Checklist:**
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/internal/k8s/builder/builder.go
+++ b/internal/k8s/builder/builder.go
@@ -26,7 +26,7 @@ func init() {
 }
 
 const (
-	defaultEnvoyImage     = "envoyproxy/envoy:v1.19-latest"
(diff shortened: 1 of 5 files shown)


andrewstucki (changes requested):
Looks like you need to regenerate the fixtures, after that will approve.

```bash
GENERATE=true go test ./...
```

andrewstucki (approved):

PR #123, opened 2022-03-10
Assign appropriate IP to Gateway

### LINKED ISSUE #116: Incorrect IP is added to k8s `Gateway` resource
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request. Searching for pre-existing feature requests helps us consolidate datapoints for identical requirements into a single place, thank you!
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

---

### Overview of the Issue

The gateway pod's IP is assigned to the `Gateway` resource in k8s ([here](https://github.com/hashicorp/consul-api-gateway/blob/6fca6de1ab65d4ea94c70fe3061c615e66fe9854/internal/k8s/reconciler/gateway.go#L169-L171)); however, this is incorrect for cases where a service of type LoadBalancer, NodePort, ClusterIP is used. The IP on the Gateway resource needs to be externally available and so should be the service's external IP, the node IP, or the internal IP respectively. If no service is created, then we can continue assigning the pod's IP as we do today.

### Reproduction Steps

1. Install to Kubernetes cluster (tested on GKE) using `serviceType: LoadBalancer` in the Consul Helm chart

1. Check IP address on `Gateway` resource:

    ```shell
    $ kubectl get gateway <gatewayName> -o yaml
    ```

1. Check external IP of `Service` resource:

    ```shell
    $ kubectl get service <serviceName>
    ```

1. Observe that they do not match

### Expected behavior

IP address on `Gateway` should match the external IP of the `Service`

nathancoleman:
Fixes #116

- Assign the external IP of a `Service` to `Gateway` instead of pod IP when service type is `LoadBalancer`
- Wait to mark `Gateway` ready until `Service` is available and has the necessary exposed IP

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -186,8 +186,9 @@ func (g *K8sGateway) validateGatewayIP(ctx context.Context) error {
 			}
 		}
 	case corev1.ServiceTypeNodePort:
-		// TODO
-		break
+		// TODO NodePorts do not always behave the same. A node in standard k8s
+		//  has a public IP address; however, this is not the case in kind.
+		g.serviceReady = true
 	default:
 		return fmt.Errorf("unsupported service type: %s", service.Spec.Type)
 	}
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -185,6 +185,13 @@ func (g *K8sGateway) validateGatewayIP(ctx context.Context) error {
 			}
 		}
 	case corev1.ServiceTypeNodePort:
+		/* For serviceType: NodePort, there isn't a consistent way to guarantee access to the
+		 * service from outside the k8s cluster. For now, we're putting the IP address of the
+		 * nodes that the gateway pods are running on.
+		 * The practitioner will have to understand that they may need to port forward into the
+		 * cluster (in the case of Kind) or open firewall rules (in the case of GKE) in order to
+		 * access the gateway from outside the cluster.
+		 */
 		pod, err := g.client.PodWithLabels(ctx, utils.LabelsForGateway(g.gateway))
 		if err != nil {
 			return err
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -170,11 +170,8 @@ func (g *K8sGateway) validateGatewayIP(ctx context.Context) error {
 
 	switch service.Spec.Type {
 	case corev1.ServiceTypeLoadBalancer:
-		if len(updated.Status.LoadBalancer.Ingress) > 0 {
-			g.serviceReady = true
-		}
-
 		for _, ingress := range updated.Status.LoadBalancer.Ingress {
+			g.serviceReady = true
 			g.addresses = append(g.addresses, ingress.IP)
 		}
 	case corev1.ServiceTypeClusterIP:
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -189,6 +189,7 @@ func (g *K8sGateway) validateGatewayIP(ctx context.Context) error {
 
 		if pod == nil {
 			g.status.Scheduled.NotReconciled = errors.New("pod not found")
+			return nil
 		}
 
 		if pod.Status.HostIP != "" {
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	gw "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	"github.com/hashicorp/consul-api-gateway/internal/core"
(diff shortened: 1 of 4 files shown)


internal/k8s/reconciler/gateway.go
```diff
@@ -149,6 +155,57 @@ func (g *K8sGateway) validateListenerConflicts() {
...
+				g.serviceReady = true
+				g.addresses = append(g.addresses, ingress.IP)
+			}
+		} else if service.Spec.Type == corev1.ServiceTypeClusterIP {
+			if updated.Spec.ClusterIP != "" {
+				g.serviceReady = true
+				g.addresses = append(g.addresses, updated.Spec.ClusterIP)
+			}
+		}
+	case corev1.ServiceTypeNodePort:
```
nathancoleman, line 187: For `serviceType: NodePort`, there isn't a great consistent way to guarantee access to the service from outside the k8s cluster. For now, we're putting the IP address of the nodes that the gateway pods are running on. The practitioner will have to understand that they might have to port forward into the cluster (in the case of Kind) or open firewall rules (in the case of GKE) in order to access the gateway from outside the cluster.

internal/k8s/reconciler/gateway.go
```diff
@@ -275,6 +328,9 @@ func (g *K8sGateway) isEqual(other *K8sGateway) bool {
 	if g.podReady != other.podReady {
 		return false
 	}
+	if g.serviceReady != other.serviceReady {
+		return false
+	}
```
nathancoleman, line 333: In addition to waiting for pods to be ready - which we were already doing - we now wait for the `Service` to be available and have an accessible IP before we mark the `Gateway` ready.

internal/k8s/reconciler/gateway.go
```diff
@@ -330,15 +386,14 @@ func (g *K8sGateway) Status() gw.GatewayStatus {
...
-	addresses := []gw.GatewayAddress{}
+	addresses := make([]gw.GatewayAddress, 0, len(g.addresses))
 	for _, address := range g.addresses {
 		addresses = append(addresses, gw.GatewayAddress{
 			Type:  &ipType,
 			Value: address,
 		})
 	}
 
-	// TODO: set addresses based off of pod/service lookup
```
nathancoleman, line 341: I _believe_ this PR addresses this TODO, but lemme know if you don't think it does.

internal/k8s/reconciler/gateway_test.go
```diff
@@ -47,20 +48,20 @@ func TestGatewayValidate(t *testing.T) {
...
 		Config: apigwv1alpha1.GatewayClassConfig{
 			Spec: apigwv1alpha1.GatewayClassConfigSpec{
-				ServiceType: &serviceType,
+				ServiceType: serviceType(core.ServiceTypeNodePort),
 			},
 		},
 	})
 	client.EXPECT().GetSecret(gomock.Any(), gomock.Any()).Return(nil, nil)
-	client.EXPECT().PodWithLabels(gomock.Any(), gomock.Any()).Return(nil, nil)
+	client.EXPECT().PodWithLabels(gomock.Any(), gomock.Any()).Return(nil, nil).Times(2)
```
nathancoleman, line 56: `PodWithLabels` now also gets called by the `Service`/IP validation process

internal/k8s/reconciler/gateway.go
```diff
@@ -166,10 +223,6 @@ func (g *K8sGateway) validatePodConditions(pod *corev1.Pod) {
 		return
 	}
 
-	if pod.Status.PodIP != "" {
```
andrewstucki, line 169: So, with this change and the `switch` above that only looks at service types, we'd no longer be populating the `addresses` field of the gateway status if no service type was defined. I'd personally prefer we just move this logic into a helper function and call it in the `if service == nil { ... }` block in the new `validateGatewayIP` function. Otherwise gateways without service-level exposure aren't going to carry information about how we can access them.

internal/k8s/reconciler/gateway.go
andrewstucki: Yep, I had actually forgotten I had thrown this comment in here back in October, so I guess I realized the shortcomings of the current approach back then but amidst additional implementations and prepping for releases totally forgot about this 😅

internal/k8s/reconciler/gateway.go
andrewstucki: Yeah, I think that's a good call since, esp in the case of something like a LoadBalancer-fronted implementation things aren't really ready until the service can be used.

internal/k8s/reconciler/gateway.go
andrewstucki: Can we actually add this as a comment in this branch of the `switch` just so we know, in code, why we're using the `HostIP` status rather than (here a potentially empty) IP from the service itself?

internal/k8s/reconciler/manager_test.go
```diff
@@ -89,7 +90,7 @@ func TestUpsertGateway(t *testing.T) {
 	client.EXPECT().PodWithLabels(gomock.Any(), gomock.Any()).Return(nil, expected)
 	require.Equal(t, expected, manager.UpsertGateway(context.Background(), inner))
 
-	client.EXPECT().PodWithLabels(gomock.Any(), gomock.Any()).Return(nil, nil)
+	client.EXPECT().PodWithLabels(gomock.Any(), gomock.Any()).Return(nil, nil).Times(2)
```
andrewstucki, line 93: mocking at its finest 😭 -- sorry for that

andrewstucki (approved):
LGTM

PR #124, opened 2022-03-11
Add support for CONSUL_LICENSE_PATH env variable

sarahalsmiller:
Added support for CONSUL_LICENSE_PATH

### How I've tested this PR:
-running go test ./...
-ci/cd testing suite

- running go test ./..., ideally with CONSUL_LICENSE_PATH and CONSUL_LICENSE set
- testing suite 

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/testing/e2e/consul.go
+++ b/internal/testing/e2e/consul.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"log"
 	"net"
 	"os"
@@ -26,10 +27,11 @@ import (
 )
 
 const (
-	defaultConsulImage            = "hashicorp/consul:1.11.4"
-	envvarConsulImage             = envvarPrefix + "CONSUL_IMAGE"
-	envvarConsulEnterpriseLicense = "CONSUL_LICENSE"
-	configTemplateString          = `
+	defaultConsulImage                = "hashicorp/consul:1.11.4"
+	envvarConsulImage                 = envvarPrefix + "CONSUL_IMAGE"
+	envvarConsulEnterpriseLicense     = "CONSUL_LICENSE"
+	envvarConsulEnterpriseLicensePath = "CONSUL_LICENSE_PATH"
+	configTemplateString              = `
 {
 	"log_level": "trace",
   "acl": {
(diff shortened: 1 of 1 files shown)


internal/testing/e2e/licensemanager/license.go
```diff
@@ -0,0 +1,39 @@
+//NOTE - this code is copied directly from this file https://github.com/hashicorp/consul-enterprise/blob/6099c553634223a9a7ddf2263d90eeaef98c50e0/agent/licensemanager/loader.go#L24-L36
+// the package isn't being used directly as part of a conscious choice to limit using consul-enterprise as a direct dependency
+
```
andrewstucki, line 3: I'm actually fine dropping this comment, it's not going to make sense to any non-internal developers hacking on this and things are simple enough that I don't mind drift in this code.

internal/testing/e2e/licensemanager/license.go
```diff
@@ -0,0 +1,39 @@
...
+package licensemanager
```
andrewstucki, line 4: Since this is super Consul-specific and we really have no other place in our testing environment that requires reading Consul licenses for the time being, I think it'd be fine just dropping the `LoadLicense` function into an unexported helper in `internal/testing/e2e/consul.go`.

If we ever needed to do anything more complicated like live-reloading, etc. it'd make sense to move it into its own subpackage, but this is simple enough I think we should just keep it with the other bootstrapping code for spinning up the Consul test environment -- makes the code easier to track.

andrewstucki, line 4: Since this is super Consul-specific and we really have no other place in our testing environment that requires reading Consul licenses for the time being, I think it'd be fine just dropping the `LoadLicense` function into an unexported helper in `internal/testing/e2e/consul.go`.

If we ever needed to do anything more complicated like live-reloading, etc. it'd make sense to move it into its own subpackage, but this is simple enough I think we should just keep it with the other bootstrapping code for spinning up the Consul test environment -- makes the code easier to track.

andrewstucki (changes requested):
Quick changes around adding this as a subpackage v. just dropping the helper into the existing code, but overall seems fine to me.

internal/testing/e2e/consul.go
```diff
@@ -635,3 +642,32 @@ func adminPolicy() *api.ACLPolicy {
 `,
 	}
 }
+
+const (
+	LicenseEnvVar     = "CONSUL_LICENSE"
+	LicensePathEnvVar = "CONSUL_LICENSE_PATH"
+)
```
andrewstucki, line 649: these constants should get dropped

andrewstucki (approved):
LGTM, please remove the stray `const` values before merging, but otherwise 👍

PR #127, opened 2022-03-15
Use ErrorOrNil for asserting no errors in multierror group

nathancoleman:
Something unexpected that I discovered while using the go-multierror library in a personal project:

Use `ErrorOrNil()` when asserting no errors from `multierror.Group`.

See docstring for `ErrorOrNil()` [here](https://github.com/hashicorp/go-multierror/blob/9974e9ec57696378079ecc3accd3d6f29401b3a0/multierror.go#L24-L28).

### How I've tested this PR:
This test fails every time
```go
func TestMultierrorGroupWait(t *testing.T) {
    var g multierror.Group
    g.Go(func() error { return nil })
    require.NoError(t, g.Wait())
}
```

This one behaves as you'd expect
```go
func TestMultierrorGroupWait(t *testing.T) {
    var g multierror.Group
    g.Go(func() error { return nil })
    require.NoError(t, g.Wait().ErrorOrNil())
}
```

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/store/memory/store.go
+++ b/internal/store/memory/store.go
@@ -106,7 +106,7 @@ func (s *Store) syncGateways(ctx context.Context, gateways ...*gatewayState) err
 			return s.syncGateway(ctx, gateway)
 		})
 	}
-	if err := syncGroup.Wait(); err != nil {
+	if err := syncGroup.Wait().ErrorOrNil(); err != nil {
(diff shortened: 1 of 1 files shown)


andrewstucki (approved):
😳

PR #128, opened 2022-03-16
Simplify comparison of gateways to use only resource version

nathancoleman:
We currently wind up with a stale gateway in memory that can't be written to the store when we want to update it.

```logs
2022-03-15T17:56:05.165Z [ERROR] memory/store.go:110: consul-api-gateway-server.state: error syncing gateways:
  error=
  | 1 error occurred:
  | 	* Operation cannot be fulfilled on gateways.gateway.networking.k8s.io "backend-namespaces": the object has been modified; please apply your changes to the latest version and try again
  | 
```

Due to this, the new route that's being added is unable to bind and thus the conformance tests fail.

After adding some code using [go-test/diff](https://github.com/go-test/deep), I was able to determine the diff between the gateway `Listeners()` during the sync process:

```diff
slice[0].gateway.ObjectMeta.Annotations.map[api-gateway.consul.hashicorp.com/config]:

-{\"serviceType\":\"LoadBalancer\",\"consul\":{\"authentication\":{},\"scheme\":\"https\",\"ports\":{\"http\":8501,\"grpc\":8502}},\"image\":{\"consulAPIGateway\":\"gcr.io/hc-9b66efef4c6f4f9a85bcdbe92a9/consul-api-gateway-oss:16\",\"envoy\":\"envoyproxy/envoy-alpine:v1.20.2\"},\"copyAnnotations\":{},\"logLevel\":\"info\"}
+{\"serviceType\":\"LoadBalancer\",\"consul\":{\"authentication\":{},\"scheme\":\"https\",\"ports\":{\"http\":8501,\"grpc\":8502}},\"image\":{\"consulAPIGateway\":\"gcr.io/hc-9b66efef4c6f4f9a85bcdbe92a9/consul-api-gateway-oss:17\",\"envoy\":\"envoyproxy/envoy-alpine:v1.20.2\"},\"copyAnnotations\":{},\"logLevel\":\"info\"}

slice[0].routeCount:
-1
+0
```

When the sync fails, it retries a couple of times and then goes silent. The gateway then never has the route listed in attached routes. Andrew mentioned that he doesn't think we want to dirty-check the route count, and I think we want the annotation config to be a snapshot of creation time. I'm not sure which of these is preventing the sync; however, including `gateway.Listeners()` in the `isEqual` comparison dirty-checks both implicitly and fixes the issue in my testing.

When determining if two gateways are equal during the sync process, only compare the resource version. This will prevent the list of things to compare from continually growing and requiring tons of mental overhead to figure out any bugs.

### How I've tested this PR:
Running the upstream conformance tests against `main` will trigger the issue above. The only fix I've found is to delete the controller pod so that the stale gateway is dumped.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -388,6 +388,9 @@ func (g *K8sGateway) isEqual(other *K8sGateway) bool {
 	if !reflect.DeepEqual(g.certificates(), other.certificates()) {
 		return false
 	}
+	if !reflect.DeepEqual(g.Listeners(), other.Listeners()) {
+		return false
+	}
 	if !conditionEqual(g.status.Scheduled.Condition(g.gateway.Generation), other.status.Scheduled.Condition(g.gateway.Generation)) {
 		return false
 	}
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -365,7 +365,7 @@ func (g *K8sGateway) ShouldUpdate(other store.Gateway) bool {
 		return false
 	}
 
-	return utils.ResourceVersionLesser(g.gateway.ResourceVersion, otherGateway.gateway.ResourceVersion)
+	return !utils.ResourceVersionGreater(g.gateway.ResourceVersion, otherGateway.gateway.ResourceVersion)
 }
 
 func (g *K8sGateway) ShouldBind(route store.Route) bool {
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -385,6 +385,9 @@ func (g *K8sGateway) isEqual(other *K8sGateway) bool {
 	if !reflect.DeepEqual(g.certificates(), other.certificates()) {
 		return false
 	}
+	if !reflect.DeepEqual(g.Listeners(), other.Listeners()) {
+		return false
+	}
 	if !conditionEqual(g.status.Scheduled.Condition(g.gateway.Generation), other.status.Scheduled.Condition(g.gateway.Generation)) {
 		return false
 	}
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -374,6 +374,9 @@ func (g *K8sGateway) Compare(other store.Gateway) store.CompareResult {
 }
 
 func (g *K8sGateway) isEqual(other *K8sGateway) bool {
+	if !reflect.DeepEqual(g.gateway.Annotations, other.gateway.Annotations) {
+		return false
+	}
 	if !reflect.DeepEqual(g.gateway.Spec, other.gateway.Spec) {
 		return false
 	}
@@ -383,9 +386,6 @@ func (g *K8sGateway) isEqual(other *K8sGateway) bool {
 
 	// check other things that may affect the pending status updates
 	if !reflect.DeepEqual(g.certificates(), other.certificates()) {
-		return false
-	}
-	if !reflect.DeepEqual(g.Listeners(), other.Listeners()) {
 		return false
 	}
 	if !conditionEqual(g.status.Scheduled.Condition(g.gateway.Generation), other.status.Scheduled.Condition(g.gateway.Generation)) {
--- a/internal/k8s/reconciler/gateway.go
+++ b/internal/k8s/reconciler/gateway.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


nathancoleman:
Though this change fixes the issue that I've been seeing, I'm actually thinking we might need to expand [this code](https://github.com/hashicorp/consul-api-gateway/blob/55ccbe65d3d9b60110670a075fd212a15958f04a/internal/k8s/controllers/gateway_class_config_controller.go#L68-L75) to evict `Gateways` using the `GatewayClasses` that we're evicting. Seems sort of hairy to keep expanding that out, and I'm not sure that change makes logical sense.

Expanding that code seems possibly nonsensical because I believe we're supposed to be able to hang onto the original config of a `Gateway`, even if the `GatewayConfigClass` that it's referencing has changed over time.

Thoughts here @andrewstucki @mikemorris @sarahalsmiller ?

mikemorris:
I'm slightly confused at when/how the gateway annotations would have been modified - was that something you did manually? My understanding is that the GatewayClass config of a Gateway should be an immutable snapshot from when the gateway was created?

nathancoleman:
The setup is that the Gateway(s) already exist, and then I do a `helm upgrade consul` (simplified) that changes `apiGateway.image` in `values.yaml`.

My understanding matches yours, that the config should be a snapshot from time of creation; however, we also still need to be able to bind new routes to that gateway. How do we bind new routes if the Gateway upsert returns the error in the PR body?

nathancoleman:
Closing for now while I check out the simplified approach to dirty-checking that we discussed a bit ago

internal/k8s/reconciler/gateway.go
```diff
@@ -352,53 +351,21 @@ func (g *K8sGateway) Listeners() []store.Listener {
...
-	if !reflect.DeepEqual(g.addresses, other.addresses) {
+
+	otherGateway, ok := other.(*K8sGateway)
+	if !ok {
 		return false
 	}
 
-	return true
+	return utils.ResourceVersionLesser(g.gateway.ResourceVersion, otherGateway.gateway.ResourceVersion)
 }
```
nathancoleman, line 369: We discussed today that, rather than having this list of comparisons continue to grow over time, we should be able to compare just the `ResourceVersion` and, if we aren't outdated, refrain from updating.

I've tested the simplified comparison using the upstream conformance tests and everything seems to work correctly.

andrewstucki: So, I think we actually want to do a `<=` check here rather than just `<` -- this is to ensure that when child resources (i.e. services/deployments) wind up triggering a reconciliation rather than a `Gateway` update, we're still able to go ahead and do the Gateway's status updates which are derived from child resources.

In the case of the child resource triggering the reconciliation, the `Gateway`'s version isn't going to change, but potentially the derived status of it may need an update, so we can't just skip the update path -- though we'll need to lean heavily on the `Status` dirty checking here:

https://github.com/hashicorp/consul-api-gateway/blob/55ccbe65d3d9b60110670a075fd212a15958f04a/internal/k8s/reconciler/gateway.go#L482-L494

to make sure we're not caught in an infinite update loop.

andrewstucki (changes requested):
Some thoughts about the comparison check here, also WDYT about updating the `Route` code and dropping the multi-state `store.CompareResult` business?

https://github.com/hashicorp/consul-api-gateway/blob/55ccbe65d3d9b60110670a075fd212a15958f04a/internal/k8s/reconciler/route.go#L261

Want to handle that here or in a follow-up PR?

internal/k8s/reconciler/gateway.go
andrewstucki: For anyone else following, the following resources can all trigger the reconciliation and may require an update to the `Gateway`'s status:

https://github.com/hashicorp/consul-api-gateway/blob/55ccbe65d3d9b60110670a075fd212a15958f04a/internal/k8s/controllers/gateway_controller.go#L84-L92

i.e. `Gateway`, `Service`, `Deployment`, `Pod`, and `ServiceAccount` (for ensuring the existence of Gateway-owned service accounts in a setup that's leveraging managed Kubernetes auth methods for Consul authentication on the deployed gateways)

nathancoleman:
@andrewstucki I noticed they're very similar and felt the itch to change the route logic as well; however, I'm planning to do that in a followup PR to keep this one from growing in scope

internal/store/memory/store.go
```diff
@@ -219,11 +219,7 @@ func (s *Store) UpsertGateway(ctx context.Context, gateway store.Gateway) error
 
 	current, found := s.gateways[id]
 
-	switch current.Compare(gateway) {
-	case store.CompareResultInvalid, store.CompareResultNewer:
-		// we have an invalid or old route, ignore it
-		return nil
-	case store.CompareResultNotEqual:
+	if current.ShouldUpdate(gateway) {
```
andrewstucki, line 222: Think we need to vary this logic a slight amount -- after this block there's still a call to `sync` gateways. Though I'm pretty sure the `sync` calls will just no-op since nothing has been updated, to be in accordance with the  previous logic we should do something like:

```go
if !current.ShouldUpdate(gateway) {
  // no-op
  return nil
}
```

nathancoleman:
Confirmed w/ trace logging that I can exercise all of the conformance tests without hammering the kube API

andrewstucki (approved):
Overall this looks vastly improved to me and much better than the runaround that is tracking down all potential fields we've failed to dirty check. Thanks for taking care of it.

PR #129, opened 2022-03-16
Facilitate running of conformance tests

nathancoleman:
In order for the upstream conformance tests to work, we need a couple of small patches applied to each `Deployment` and a `ProxyDefaults` resource created.

Long-term, I'd like to run without these patches; however, they're required for now as no `containerPort` is declared on the upstream `Deployments` and the test suite uses `http` protocol where Consul defaults to `tcp`.

### How I've tested this PR:
See included README instructions

_Note: The change in #128 is required for the conformance tests to work at all right now._

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


andrewstucki (approved):
As this is just mostly documentation + the series of necessary `kustomize` patches I'm fine adding this. Since the `run.sh` script though requires modification between runs to apply the proper test manifests think maybe we should just drop it for the time being and stick with the manual instructions?

PR #131, opened 2022-03-18
Use Generated RBAC policy for tests

sarahalsmiller:
-Small fix to avoid kubectl exec error from being swallowed
- Updated helper functions to create a service account of an arbitrary name
- Updated createClusterRole to take in a path to k8s yaml file with a clusterrole definition, or an array of rbac.PolicyRules
- Updated gateway.go to load a service account client instead of the default

### How I've tested this PR:
- `go test -tags e2e ./internal/commands/server`
- CI testing pipeline
- CI pipeline pases
- go test -tags e2e ./internal/commands/server

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/testing/e2e/gateway.go
```diff
@@ -51,10 +52,15 @@ func (p *gatewayTestEnvironment) run(ctx context.Context, namespace string, cfg
 	}
 	k8sSecretClient.AddToMultiClient(secretClient)
```
andrewstucki, line 54: You'll also want to change the initialization of the `secretClient` to leverage the scoped client

PR #132, opened 2022-03-21
Simplify naming of sync adapter in consideration of pkg name

nathancoleman:
Since the `ConsulSyncAdapter` resides in the `consul` package, we can simplify the name to `SyncAdapter` which makes things less verbose while conveying the same thing.

### How I've tested this PR:
CI

CI

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/adapters/consul/sync.go
+++ b/internal/adapters/consul/sync.go
@@ -21,7 +21,7 @@ type syncState struct {
 	defaults  *consul.ConfigEntryIndex
 }
 
-type ConsulSyncAdapter struct {
+type SyncAdapter struct {
 	logger hclog.Logger
 	consul *api.Client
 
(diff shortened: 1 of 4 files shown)


mikemorris (commented):
🤔 Guessing we should probably follow a consistent naming pattern for interface implementations - I believe `ConsulSyncAdapter` implements [`SyncAdapter`](https://github.com/hashicorp/consul-api-gateway/blob/main/internal/core/interfaces.go#L7-L12), similarly to how [`K8sListener`](https://github.com/hashicorp/consul-api-gateway/blob/main/internal/k8s/reconciler/listener.go) implemented the resolved [Listener](https://github.com/hashicorp/consul-api-gateway/blob/main/internal/store/interfaces.go#L61-L68) interface.

andrewstucki (approved):

nathancoleman:
@mikemorris I don't think we need to worry about pairing up structs and interfaces with naming conventions since you can often have structs that fulfill multiple interfaces and are very loosely coupled / not coupled at all. The `K8sListener` also doesn't hit the antipattern of having the package name in the struct name

PR #133, opened 2022-03-21
Add docstrings + test coverage for route merging

nathancoleman:
As I'm working on better understanding how our controller interacts w/ Consul, I thought I'd add some docstrings and naming changes that helped me understand the code

I'd like for others to read through the docstrings and make sure that my understanding was correct as well as confirming whether the added docstrings are helpful.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/adapters/consul/http.go
+++ b/internal/adapters/consul/http.go
@@ -251,36 +251,35 @@ func httpServiceDefault(entry api.ConfigEntry, meta map[string]string) *api.Serv
 	}
 }
 
-type flattenedRoute struct {
+type hostnameMatch struct {
 	match    core.HTTPMatch
 	filters  []core.HTTPFilter
 	services []core.HTTPService
 }
 
-type flattenedRouteMap struct {
-	flattenedRoutesByHostname map[string][]flattenedRoute
+type routeConsolidator struct {
+	matchesByHostname map[string][]hostnameMatch
 }
 
-func newFlattenedRouteMap() *flattenedRouteMap {
-	return &flattenedRouteMap{
-		flattenedRoutesByHostname: map[string][]flattenedRoute{},
+func newRouteConsolidator() *routeConsolidator {
+	return &routeConsolidator{
+		matchesByHostname: map[string][]hostnameMatch{},
 	}
 }
 
-// flatten takes a new route and flattens its routing rules out per hostname.
+// addRoute takes a new route and flattens its rule matches out per hostname.
 // This is required since a single route can specify multiple hostnames, and a
 // single hostname can be specified in multiple routes. Routing for a given
 // hostname must behave based on the aggregate of all rules that apply to it.
-func (f *flattenedRouteMap) flatten(route core.HTTPRoute) {
+func (f *routeConsolidator) addRoute(route core.HTTPRoute) {
 	for _, host := range route.Hostnames {
-		routesForHost, ok := f.flattenedRoutesByHostname[host]
+		matches, ok := f.matchesByHostname[host]
 		if !ok {
-			routesForHost = []flattenedRoute{}
+			matches = []hostnameMatch{}
 		}
 
-		// Add one route for each match of each rule
 		for _, rule := range route.Rules {
-			// If a rule has no matches defined, add default match
+			// If a rule has no matches defined, addRoute default match
 			if len(rule.Matches) == 0 {
 				rule.Matches = []core.HTTPMatch{{
 					Path: core.HTTPPathMatch{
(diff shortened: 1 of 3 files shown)


internal/adapters/consul/http.go
```diff
@@ -219,18 +221,17 @@ func hostsKey(hosts ...string) string {
...
+// compareHTTPRules implements some of the order of precedence for routes specified by the K8s Gateway API spec.
+// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteRule
+//
+// Ordering prefers matches based on the largest number of:
+//
+//   1. characters in a matching non-wildcard hostname
+//   2. characters in a matching hostname
+//   3. characters in a matching path
+//   4. header matches
+//   5. query param matches
```
nathancoleman, line 233: Added a link to the spec where it covers this order of precedence. I also noticed that we don't apply the first and second comparisons here. I'm assuming they're applied elsewhere in the codebase?

internal/adapters/consul/http.go
```diff
@@ -281,42 +289,54 @@ func (f *flattenedRouteMap) flatten(route core.HTTPRoute) {
...
-				Matches:  []core.HTTPMatch{match.match},
-				Filters:  match.filters,
-				Services: match.services,
-			})
-		}
-		name := gateway.ID.Service + "-" + hostsKey(hostname)
-		coreRoutes = append(coreRoutes, core.HTTPRoute{
+	var routes []core.HTTPRoute
+
+	for hostname, rules := range f.flattenedRoutesByHostname {
```
nathancoleman, line 312: I'm actually wondering if, given the way that this logic all works, it might make sense to change `flattenedRouteByHostname` to `rulesByHostname` or something similar. Thoughts?

nathancoleman (commented):
Personal review

internal/adapters/consul/http.go
andrewstucki: Nope, it's all internal to `envoy` and AFAIK we have 0 control over it (though envoy's prioritization does match domains this way). I've never found documentation on this, but here are some links that we might consider adding as part of the comment:

- Comment in [Istio](https://github.com/istio/istio/issues/35826#issuecomment-957184380) repo
- Comment in [envoy](https://github.com/envoyproxy/envoy/issues/18874#issuecomment-958430741) repo
- I'm pretty sure this is the [matching code in envoy](https://github.com/envoyproxy/envoy/blob/5c4d4bd957f9402eca80bef82e7cc3ae714e04b4/source/common/router/config_impl.cc#L1645-L1682)

nathancoleman: Thanks! I added the third link which seemed the most directly related and least ephemeral of the three.

internal/adapters/consul/http.go
```diff
@@ -281,42 +291,54 @@ func (f *flattenedRouteMap) flatten(route core.HTTPRoute) {
...
-				Matches:  []core.HTTPMatch{match.match},
-				Filters:  match.filters,
-				Services: match.services,
-			})
-		}
-		name := gateway.ID.Service + "-" + hostsKey(hostname)
-		coreRoutes = append(coreRoutes, core.HTTPRoute{
+// consolidateRoutes combines all rules into the shortest possible list of routes
+// with one route per hostname containing all rules for that hostname.
+func (f *routeConsolidator) consolidateRoutes(gateway core.ResolvedGateway) []core.HTTPRoute {
```
andrewstucki, line 311: Since we're already in the process of simplifying some of the naming conventions 😄 

```suggestion
func (f *routeConsolidator) consolidate(gateway core.ResolvedGateway) []core.HTTPRoute {
```

internal/adapters/consul/http.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/consul/api"
 )
 
-// httpRouteToServiceDiscoChain will convert a k8s HTTPRoute to a Consul service-router config entry and 0 or
+// httpRouteDiscoveryChain will convert a k8s HTTPRoute to a Consul service-router config entry and 0 or
```
andrewstucki, line 13: 👍 I believe this was originally written as `DiscoChain` since it's a well-known internal nickname for discovery chains in core, but I much prefer the explicit name, and that way it doesn't make me think about these guys:

Bee Gees

andrewstucki (approved):
One small suggestion on the `consolidateRoutes` name, other than that, LGTM

PR #137, opened 2022-03-24
[WIP] Initial Serialization Groundwork

andrewstucki:

So this refactor is an initial attempt to slim down much of the logic in our Gateway/Listener and Route wrappers. The main goal is to get to the point where we can actually fully Marshal and Unmarshal our stateful wrappers. This is to begin the process for introducing additional durable storage backends for configuration as we expand past the Kubernetes world.

Currently we have things like loggers, http clients, and all other sorts of non-serializable fields on our wrappers. This is an initial attempt to:

1. Remove some of the non-serializable fields from the wrappers
2. Slim down so we can eventually remove the `Listener` wrapper since it should just be a part of a `Gateway`
3. Remove some of the interface methods in the store objects to further simplify the wrappers
4. Introduce `Binder` to remove much of the complex binding logic from the listener and
5. Introduce `Factory` to help with initializing routes/gateways and eventually act as a serialization interface for marshaling/unmarshaling and rehydrating those same objects


nathancoleman:
@andrewstucki are we at a point where we can close this in favor of
- #160 
- #163 
- #165

PR #140, opened 2022-04-05
Address linting failures

nathancoleman:
- Changes to please the linter, no functional impact
- Show all linting failures on pull requests

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         args: |
           --verbose
         # Optional: show only new issues if it's a pull request. The default value is `false`.
-        only-new-issues: true
+        only-new-issues: false
         # Optional: if set to true then the action don't cache or restore ~/go/pkg.
         skip-pkg-cache: true
         # Optional: if set to true then the action don't cache or restore ~/.cache/go-build.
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         args: |
           --verbose
         # Optional: show only new issues if it's a pull request. The default value is `false`.
-        only-new-issues: true
+        only-new-issues: false
```
nathancoleman, line 28: Now that all linting issues are addressed, my thought is that any new ones would be introduced by a PR.

Prefer having this flag set to `false` so we avoid PRs passing and then finding surprises when CI fails on `main` post-merge.

internal/envoy/secrets.go
```diff
@@ -221,15 +221,15 @@ func (s *secretManager) watch(ctx context.Context, names []string, node string)
...
+	err := s.updateCertificates(certificates)
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
 // UnwatchAll is used to completely unwatch all a node's secrets
-func (s *secretManager) UnwatchAll(ctx context.Context, node string) error {
+func (s *secretManager) UnwatchAll(_ context.Context, node string) error {
```
nathancoleman, line 232: I left these here instead of removing because modifying the interface that this fulfills seemed like a bigger change than I wanted to make based on the linter today

nathancoleman (commented):
Personal Review

andrewstucki (approved):

PR #142, opened 2022-04-05
Validate ReferencePolicy for cross-namespace routing, log future enforcement

nathancoleman:
- Update RBAC so that the controller can retrieve `ReferencePolicies` defined in the kube-sig Gateway API
- Validate `ReferencePolicy` for cross-namespace routing
- **Only log** if the route would be considered invalid in the future
- Add TODO + FUTURE comments as appropriate for followup work to actually enforce the `ReferencePolicy` requirement

Note: Followup PRs will handle setting the status on the route appropriately and handling updates to the `ReferencePolicy` itself.

### How I've tested this PR:
- 🤖 tests pass, including added test coverage
- Ran Mike's new conformance test coverage from https://github.com/kubernetes-sigs/gateway-api/pull/1081 and verified that it fails in ways we expect for now
- All existing test coverage still passes which indicates that we succeeded in this PR having no functional impact

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -479,5 +479,12 @@ func (g *gatewayClient) HasManagedDeployment(ctx context.Context, gw *gateway.Ga
 }
 
 func (g *gatewayClient) GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error) {
-	return []gateway.ReferencePolicy{}, nil
+	list := &gateway.ReferencePolicyList{}
+	if err := g.Client.List(ctx, list, client.MatchingLabels(utils.LabelsByNamespace(namespace))); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, NewK8sError(err)
+	}
+	return list.Items, nil
 }
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -480,7 +480,7 @@ func (g *gatewayClient) HasManagedDeployment(ctx context.Context, gw *gateway.Ga
 
 func (g *gatewayClient) GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error) {
 	list := &gateway.ReferencePolicyList{}
-	if err := g.Client.List(ctx, list, client.MatchingLabels(utils.LabelsByNamespace(namespace))); err != nil {
+	if err := g.Client.List(ctx, list, client.InNamespace(namespace)); err != nil {
 		if k8serrors.IsNotFound(err) {
 			return nil, nil
 		}
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -477,3 +477,7 @@ func (g *gatewayClient) HasManagedDeployment(ctx context.Context, gw *gateway.Ga
 	}
 	return false, nil
 }
+
+func (g *gatewayClient) GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error) {
+	return []gateway.ReferencePolicy{}, nil
+}
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -73,7 +73,7 @@ type Client interface {
 	EnsureServiceAccount(ctx context.Context, owner *gateway.Gateway, serviceAccount *core.ServiceAccount) error
 
 	//referencepolicy
-	GetReferencePoliciesInNamespace(ctx context.Context, namespace string) (*[]gateway.ReferencePolicy, error)
+	GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error)
 }
 
 type gatewayClient struct {
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -472,18 +472,12 @@ func (g *gatewayClient) HasManagedDeployment(ctx context.Context, gw *gateway.Ga
 	if err := g.Client.List(ctx, list, client.MatchingLabels(utils.LabelsForGateway(gw))); err != nil {
 		return false, NewK8sError(err)
 	}
-	if len(list.Items) > 0 {
-		return true, nil
-	}
-	return false, nil
+	return len(list.Items) > 0, nil
 }
 
 func (g *gatewayClient) GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error) {
 	list := &gateway.ReferencePolicyList{}
 	if err := g.Client.List(ctx, list, client.InNamespace(namespace)); err != nil {
-		if k8serrors.IsNotFound(err) {
-			return nil, nil
-		}
 		return nil, NewK8sError(err)
 	}
 	return list.Items, nil
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -71,6 +71,9 @@ type Client interface {
 	CreateOrUpdateService(ctx context.Context, service *core.Service, mutators ...func() error) (bool, error)
 	DeleteService(ctx context.Context, service *core.Service) error
 	EnsureServiceAccount(ctx context.Context, owner *gateway.Gateway, serviceAccount *core.ServiceAccount) error
+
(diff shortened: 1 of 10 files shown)


internal/k8s/reconciler/utils.go
```diff
@@ -119,6 +120,61 @@ func routeAllowedForListenerNamespaces(ctx context.Context, gatewayNS string, al
 	return false, nil
 }
 
+// routeAllowedForHTTPBackend determines whether the route is allowed
+// for the backend either by being in the same namespace or by having
+// an applicable ReferencePolicy in the same namespace as the backend.
+//
+// TODO This func is currently called once for each backendRef on a route and results
+//   in fetching ReferencePolicies more than we technically have to in some cases
+func routeAllowedForHTTPBackend(ctx context.Context, route *gw.HTTPRoute, backendRef gw.HTTPBackendRef, c gatewayclient.Client) (bool, error) {
```
andrewstucki, line 129: Since it looks like there's nothing `HTTPRoute` specific here (only `GetNamespace()` and `GroupVersionKind()` called) I'm assuming you're going to modify this to be `route Route` instead of `route *gw.HTTPRoute` and `backendRef gw.BackendRef` instead of `backendRef gw.HTTPBackendRef` (`HTTPBackendRef` [embeds](https://github.com/kubernetes-sigs/gateway-api/blob/ef302d4b80d097f7b072a506adcf01b33194f27c/apis/v1alpha2/httproute_types.go#L841) `BackendRef`) so that we can have this be a generic `ReferencePolicy` implementation?

internal/k8s/reconciler/route.go
```diff
@@ -362,8 +363,15 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 		for _, httpRule := range route.Spec.Rules {
 			rule := httpRule
 			routeRule := service.NewRouteRule(&rule)
+
 			for _, backendRef := range rule.BackendRefs {
 				ref := backendRef
+
+				allowed, err := routeAllowedForHTTPBackend(ctx, route, ref, r.client)
+				if err != nil || !allowed {
+					return fmt.Errorf("route not allowed for backend %q", ref.Name)
```
andrewstucki, line 372: I know it's reserved for follow-up, but I'd actually probably prefer just no-op'ing in this branch if we're not going to set the `Status` in this commit, namely because this is going to trigger the reconciliation loop to re-run by throwing an error here.

RE statuses, it's not too bad what you'll need to do -- basically just add an entry for `RefNotPermitted` as per [this part of the spec](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.HTTPBackendRef) to `statuses.yaml` here:

https://github.com/hashicorp/consul-api-gateway/blob/8ff752ad0295df4f3daee27b5c67b82f42b75694/internal/k8s/reconciler/config/statuses.yaml#L129

Regenerate the status structures with `go generate ./...` and you should be able to do something like `status.ResolvedRefs.RefNoPermitted = fmt.Errorf("route not allowed for backend %q", ref.Name)`

The one caveat to this and what makes status values in the API spec really wonky (and we may want to actually raise upstream) is that the spec is inconsistent AFAICT with regard to the particularities of `ResolvedRefs`, specifically, from the link above:

> If there is a cross-namespace reference to an existing object that is not covered by a ReferencePolicy, the controller must ensure the “ResolvedRefs” condition on the Route is set to status: False, with the “RefNotPermitted” reason and not configure this backend in the underlying implementation.

While later on there's also a value for `RefNotPermitted` in the listener status:

> This reason is used with the “ResolvedRefs” condition when one of the Listener’s Routes has a BackendRef to an object in another namespace, where the object in the other namespace does not have a ReferencePolicy explicitly allowing the reference.

So interestingly enough, there at least seems to me to be uncertainty about whether a route should be considered "bound" or not to a listener if it's missing a `ReferencePolicy` and whether or not the whole listener should then have a bad status (i.e. is this a condition for the route or for the listener?). I'm assuming that the latter is a mistake in the spec and that this properly belongs on the route and should just prevent the route from binding to the listener. Yay the complexities of statuses!

internal/k8s/reconciler/utils.go
nathancoleman: Getting there with the next set of items on the checklist above. I usually work by starting with the specific thing and then generalizing

internal/k8s/reconciler/route.go
nathancoleman: logging for now instead

internal/k8s/reconciler/route.go
```diff
@@ -368,12 +368,11 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
...
 				ref := backendRef
 
-				logger := r.logger.With("backendRef", ref.Name, "namespace", ref.Namespace)
-				logger.Warn("testing if route is allowed for backend ref")
-
 				allowed, err := routeAllowedForBackendRef(ctx, r.Route, ref.BackendRef, r.client)
 				if err != nil || !allowed {
-					logger.Error("route not allowed")
+					// FUTURE Set appropriate status
+					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
```
andrewstucki, line 374: 💯

internal/k8s/reconciler/route.go
```diff
@@ -360,10 +361,20 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
...
 			routeRule := service.NewRouteRule(&rule)
+
 			for _, backendRef := range rule.BackendRefs {
 				ref := backendRef
+
+				allowed, err := routeAllowedForBackendRef(ctx, r.Route, ref.BackendRef, r.client)
+				if err != nil || !allowed {
+					// FUTURE Set appropriate status
+					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+					continue
```
nathancoleman, line 375: Going back and forth myself over letting this fall through to the resolver below for now so we don't have any functional impact with the merge of this PR vs. early out and just not setting the status until followup PR. Thoughts @andrewstucki @mikemorris @sarahalsmiller ?

andrewstucki: Yeah, I'd just let this fall through for now if we're not going to set the status until the follow-up so that the functionality actually gets changed in a single commit -- the `continue` here  _should_ cause some silent routing failures for a bit since the route will have no errors on it but also no resolved services that it can write into Consul once it binds to a gateway.

Also, before I do a full-blown formal review, just to mention, you should probably return if `err != nil` since the only thing that could cause that in the underlying code is a Kubernetes client error (i.e. a network issue, api server issue, auth issue, etc., all of which we _do_ want to reschedule for reconciliation) -- it's only if we can actually determine that the route isn't allowed to reference tthe given `backendRefs` values that we should mark the Status as `RefNotAllowed`.

nathancoleman: Done and done

internal/k8s/gatewayclient/gatewayclient.go
```diff
@@ -474,3 +477,14 @@ func (g *gatewayClient) HasManagedDeployment(ctx context.Context, gw *gateway.Ga
 	}
 	return false, nil
 }
+
+func (g *gatewayClient) GetReferencePoliciesInNamespace(ctx context.Context, namespace string) ([]gateway.ReferencePolicy, error) {
+	list := &gateway.ReferencePolicyList{}
+	if err := g.Client.List(ctx, list, client.InNamespace(namespace)); err != nil {
+		if k8serrors.IsNotFound(err) {
```
andrewstucki, line 484: nit: we shouldn't need this conditional branch since a `List` invocation is just going to return a list with no items and shouldn't ever hand back `IsNotFound` -- so think we just need the `err != nil` check.

internal/k8s/reconciler/utils_test.go
```diff
@@ -359,3 +362,132 @@ func TestGatewayStatusEqual(t *testing.T) {
...
+							}},
+							To: []gw.ReferencePolicyTo{{
+								Group: "",
+								Kind:  "Service",
+								Name:  toName,
+							}},
+						},
+					}
+
+					throwawayPolicy := gw.ReferencePolicy{
```
andrewstucki, line 466: So, just to make sure, this here just to demonstrate we're filtering out policies via that `validFrom` check, right?

andrewstucki (approved):
Overall looks pretty 👍 to me. Quick nit is the only thing I'd change from what I can see.

internal/k8s/reconciler/utils_test.go
nathancoleman: Yep yep

internal/k8s/reconciler/route.go
mikemorris: Opened an issue upstream at https://github.com/kubernetes-sigs/gateway-api/issues/1112 about the lack of clarity in how Listener "acceptance" of a route and corresponding Route and Listener status conditions should be set.

PR #143, opened 2022-04-07
Set status error message 

sarahalsmiller:
- Add RefNotPermitted error to Resolution Errors
- Properly set status when ReferencePolicy doesn't exist

### How We've tested this PR:
- Manually running the conformance tests defined in this PR https://github.com/kubernetes-sigs/gateway-api/pull/1081

-Note that the conformance test passes
- Manually running the conformance tests if needed

### Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 


internal/k8s/reconciler/route.go
```diff
@@ -205,6 +205,8 @@ func (r *K8sRoute) OnBindFailed(err error, gateway store.Gateway) {
 				status.ResolvedRefs.ConsulServiceNotFound = err
 			case service.K8sServiceResolutionErrorType:
 				status.ResolvedRefs.ServiceNotFound = err
+			case service.RefNotPermittedErrorType:
```
andrewstucki, line 208: Gah, I totally forgot about the need to do this here -- glad you figured it out. For some context for everyone's sake if you're wondering why the code is like this, basically we can't set the status of the route directly when it has something that's invalid.

"Why" you ask? Because according to the spec, routes only have statuses in relation to the parent refs that they specify (yes, this makes route statuses all the more confusing when something is invalid for a route globally, as is the case here).

The result? We memoize any resolution errors from our validation stage onto the struct that we store and then, once binding to a gateway inevitably fails, we dump the resolution errors onto the parentRef-based status.

internal/k8s/reconciler/route.go
```diff
@@ -371,8 +373,10 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 				if err != nil {
 					return err
 				} else if !allowed {
-					// FUTURE Set appropriate status and early out
-					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching referencepolicy refName: %v refNamespace: %v", ref.Name, *ref.Namespace)
+					err := service.NewRefNotPermittedError(msg)
+					r.resolutionErrors.Add(err)
```
andrewstucki, line 378: nit:

```suggestion
					r.resolutionErrors.Add(service.NewRefNotPermittedError(msg))
```

internal/k8s/reconciler/route.go
```diff
@@ -411,8 +415,10 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 		if err != nil {
 			return err
 		} else if !allowed {
-			// FUTURE Set appropriate status and early out
-			r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+			msg := fmt.Sprintf("Cross-namespace routing not allowed without matching referencepolicy refName: %v refNamespace: %v", ref.Name, *ref.Namespace)
+			err := service.NewRefNotPermittedError(msg)
+			r.resolutionErrors.Add(err)
```
andrewstucki, line 420: same as above, no real need for the intermediate assignment and the line isn't that long if you just inline the error construction:

```suggestion
			r.resolutionErrors.Add(service.NewRefNotPermittedError(msg))
```

internal/k8s/service/resolver.go
```diff
@@ -30,6 +30,7 @@ const (
 	ConsulServiceResolutionErrorType
 	GenericResolutionErrorType
 	NoResolutionErrorType
+	RefNotPermittedErrorType
```
andrewstucki, line 33: Thinking we may want to refactor this `ResolutionErrors` struct a bit.

As of now we basically store a bunch of errors that can happen during our resolution process here:

https://github.com/hashicorp/consul-api-gateway/blob/49e3d2fa58ebdef57122d4946c7b99f507400cfe/internal/k8s/service/resolver.go#L72-L79

When we go to set the final `ResolvedRefs` status in the above `OnBindFailed`, we flatten all of the errors that we've had and wrap the flattened error strings into a single "uber error":

https://github.com/hashicorp/consul-api-gateway/blob/49e3d2fa58ebdef57122d4946c7b99f507400cfe/internal/k8s/service/resolver.go#L121-L133

Right now, as written, the `Flatten()` call will result in this branch getting hit:

https://github.com/hashicorp/consul-api-gateway/blob/49e3d2fa58ebdef57122d4946c7b99f507400cfe/internal/k8s/service/resolver.go#L126

due to the lack of `RefNotPermittedErrorType` being taken into account in the error storage on the struct. Since we determine that a ref isn't permitted before resolving any other references on the route, I'm thinking it might make sense just to store the error directly on the struct and if it's not `nil` return the error with a `return RefNotPermittedErrorType, r.refError` in the call to `Flatten()`. Another potential way of handling this is by storing the error for the time being directly on the `K8sRoute` object rather than introducing a new error type here.

If anyone else has other suggestions too I'd be fine with that.

andrewstucki (commented):
know this is still a WIP, but leaving some initial feedback that might help with implementation.

internal/k8s/reconciler/route.go
andrewstucki: Another thing to note, the reason we can't just store statuses on all parent refs for the route prior to attempting the insertion is because during the reference resolution process we don't actually know which gateways that the route references are managed by our controller -- we're not allowed to touch statuses for gateways that aren't managed by us.

internal/k8s/service/resolver.go
sarahalsmiller: I'm cool with implementing that, as long as there isn't going to be a potential situation where we might end up with more than one of them.

andrewstucki: Actually, that's a good point, since you can have multiple `backendRefs` with multiple required `ReferencePolicy` values, it does make sense to keep track of all of them and give feedback that a user could potentially be missing a few. So, I'm thinking that we can just extend `ResolutionErrors`, the storage arrays, `Add`, and the `Flatten` method. That's probably the most "correct" simplest way of doing things for now.

Thoughts @nathancoleman @mikemorris ?

nathancoleman: Agree

mikemorris:
Yep, I can work with @sarahalsmiller on writing a changelog entry using the new tooling today!

.github/workflows/conformance.yml
```diff
@@ -41,8 +41,8 @@ jobs:
       - name: Clone gateway-api
         uses: actions/checkout@v2
         with:
-          repository: "nathancoleman/gateway-api"
-          ref: "eventually-consistent-conformance"
+          repository: "hashicorp/gateway-api"
+          ref: "conformance/reference-policy-60s-timeout"
```
nathancoleman, line 45:
```suggestion
          ref: "60s-time-to-consistency"
```

I pushed a branch to our hashicorp fork that's name better indicates what it is

internal/k8s/reconciler/route_test.go
```diff
@@ -266,6 +266,15 @@ func TestRouteMergedStatusAndBinding(t *testing.T) {
...
+
+	statuses = route.MergedStatus().Parents
+	require.Len(t, statuses, 3)
+	require.Equal(t, "expected", string(statuses[0].ParentRef.Name))
+	require.Equal(t, "expected", string(statuses[0].ControllerName))
+	require.Equal(t, "refnotpermitted expected", statuses[0].Conditions[0].Message)
+	require.Equal(t, RouteConditionReasonBindError, statuses[0].Conditions[0].Reason)
+
 	// check binding for non-existent route
 	gateway = NewK8sGateway(&gw.Gateway{
```
andrewstucki, line 279: So, I was thinking something more along the lines of this:

```suggestion
	// check route ref
	route = NewK8sRoute(inner, K8sRouteConfig{
		ControllerName: "expected",
		Logger:         hclog.NewNullLogger(),
	})
	route.resolutionErrors.Add(service.NewRefNotPermittedError("not found"))
	route.OnBindFailed(nil, gateway)
	statuses = route.MergedStatus().Parents
	require.Len(t, statuses, 3)
	require.Equal(t, "expected", string(statuses[0].ParentRef.Name))
	require.Equal(t, "expected", string(statuses[0].ControllerName))
	require.Equal(t, "not found", statuses[0].Conditions[1].Message)
	require.Equal(t, RouteConditionReasonRefNotPermitted, statuses[0].Conditions[1].Reason)

	// check binding for non-existent route
	route = NewK8sRoute(inner, K8sRouteConfig{
		ControllerName: "expected",
		Logger:         hclog.NewNullLogger(),
	})
	gateway = NewK8sGateway(&gw.Gateway{
```

Also, looks like you'll need to add some additional error concatenation code here:
https://github.com/hashicorp/consul-api-gateway/blob/a4fc8cd904df96e044563e57946783c818ee4424/internal/k8s/service/resolver.go#L120

since right now there's no casing for `refNotPermittedErrors`

andrewstucki (changes requested):

internal/k8s/reconciler/route_test.go
andrewstucki: just to clarify, the above code didn't actually exercise the case/switch statement added to `OnBindFailed` since there wasn't anything in `resolutionErrors`

internal/k8s/reconciler/config/statuses.yaml
```diff
@@ -257,4 +265,4 @@
           description: >
             This reason is used with the “ResolvedRefs” condition when one of the Listener’s Routes has a BackendRef
             to an object in another namespace, where the object in the other namespace does not have a ReferencePolicy
-            explicitly allowing the reference.
\ No newline at end of file
+            explicitly allowing the reference.
```
nathancoleman: Do we actually use this to bubble up the error to the listener, or is this something we can rip out with the changes in this PR?

internal/k8s/reconciler/route.go
```diff
@@ -371,8 +373,10 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 				if err != nil {
 					return err
 				} else if !allowed {
-					// FUTURE Set appropriate status and early out
-					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching referencepolicy refName: %v refNamespace: %v", ref.Name, *ref.Namespace)
+					r.logger.Warn(msg)
```
nathancoleman, line 377: I'm used to logs getting piped into systems like Splunk where it's best to put the unique pieces of information in the metadata like this:
```go
r.logger.Warn("Cross-namespace routing not allowed without matching ReferencePolicy", "refName", ref.Name, "refNamespace", ref.Namespace)
```

I'm not sure if this is a pattern that HashiCorp uses but am curious if someone knows

internal/k8s/reconciler/route_test.go
```diff
@@ -266,6 +266,20 @@ func TestRouteMergedStatusAndBinding(t *testing.T) {
...
+	// check route ref
+	route = NewK8sRoute(inner, K8sRouteConfig{
+		ControllerName: "expected",
+		Logger:         hclog.NewNullLogger(),
+	})
+	route.resolutionErrors.Add(service.NewRefNotPermittedError("not found"))
+	route.OnBindFailed(nil, gateway)
+	statuses = route.MergedStatus().Parents
+	require.Len(t, statuses, 3)
+	require.Equal(t, "expected", string(statuses[0].ParentRef.Name))
```
nathancoleman, line 278: It's helpful to use the `assert` package whenever failing the assertion doesn't invalidate the rest of the test. In cases like this, it's helpful to know if the later assertions pass/fail even if this one fails, and `require` immediately ends the test on a failure.

internal/k8s/reconciler/route_test.go
```diff
@@ -531,11 +545,6 @@ func TestRouteValidateDontAllowCrossNamespace(t *testing.T) {
 			},
 		}, nil)
 
-	// FUTURE Remove once status is set and ReferencePolicy requirement is enforced
-	resolver.EXPECT().
-		Resolve(gomock.Any(), gomock.Any()).
-		Return(&service.ResolvedReference{Type: service.ConsulServiceReference, Reference: &service.BackendReference{}}, nil)
-
 	// FUTURE Assert appropriate status set on route and !route.IsValid() once ReferencePolicy requirement is enforced
```
nathancoleman, line 548: Is it reasonable the assert these things? Comment can be removed either way

nathancoleman (approved):
Some general preferences but nothing to block merge. Nice work!

internal/k8s/reconciler/route.go
andrewstucki: Yeah, so the `go-hclog` project we use everywhere (including for our loggers in this project) is a more or less lightweight structured logger, so we should definitely leverage that rather than constructing plain ole strings. Also, if someone turns on [`logJSON`](https://www.consul.io/docs/k8s/helm#v-global-logjson) in `helm` -- using the structured logs will move these into their own fields that can be easily filtered in a centralized log store as you mention.

andrewstucki (changes requested):
Sorry, just one more change for message clarity and to avoid a case where you'd get a panic.

internal/k8s/reconciler/route.go
```diff
@@ -371,8 +373,10 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 				if err != nil {
 					return err
 				} else if !allowed {
-					// FUTURE Set appropriate status and early out
-					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching referencepolicy refName: %v refNamespace: %v", ref.Name, *ref.Namespace)
```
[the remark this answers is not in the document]
sarahalsmiller: It actually shouldn't panic as is (unless ref became nil), because %v just should just print nil, but I'll go ahead and make a quick helper function and update the error message accordingly.

internal/k8s/reconciler/route.go
```diff
@@ -371,8 +373,10 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 				if err != nil {
 					return err
 				} else if !allowed {
-					// FUTURE Set appropriate status and early out
-					r.logger.Warn("Cross-namespace routing not allowed, will enforce in future", "refName", ref.Name, "refNamespace", ref.Namespace)
+					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching ReferencePolicy for Service %q", route.GetNamespace())
```
andrewstucki, line 376: this should say:

```suggestion
					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching ReferencePolicy for Service %q", getServiceID(ref, route.GetNamespace()))
```

as below.

andrewstucki (approved):
Last fix, but looks good once that's done, so going to approve -- thanks @sarahalsmiller

internal/k8s/reconciler/config/statuses.yaml
mikemorris: [For future reference] this has been raised upstream at https://github.com/kubernetes-sigs/gateway-api/issues/1112#issuecomment-1099655762

PR #145, opened 2022-04-12
Enable slack webhook

sarahalsmiller:
Enable slack webhook

### How I've tested this PR:
By opening this PR with the conformance test label

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -109,12 +109,13 @@ jobs:
         working-directory: "gateway-api/conformance"
         run: go test -v -timeout 10m ./ --gateway-class consul-api-gateway
 
-      # TODO Waiting for webhook approval
-      # - name: Report Status
-      #   if: always()
-      #   uses: ravsamhq/notify-slack-action@v1
-      #   with:
-      #     status: ${{ job.status }}
-      #     notify_when: 'failure'
-      #   env:
-      #     SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}
+      - name: Report Status
+        if: always() && github.ref == 'refs/heads/main'
+        uses: ravsamhq/notify-slack-action@v1
+        with:
+          status: ${{ job.status }}
+          mention_groups: '!here'
+          mention_groups_when: 'failure,warnings'
+
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}


nathancoleman:
We discussed in Slack needing helpdesk to change the target URL, right? Does it make sense to go ahead and merge this or put it on hold til helpdesk does their thing?

nathancoleman:
We discussed in Slack needing helpdesk to change the target URL, right? Does it make sense to go ahead and merge this or put it on hold til helpdesk does their thing?

sarahalsmiller:
> We discussed in Slack needing helpdesk to change the target URL, right? Does it make sense to go ahead and merge this or put it on hold til helpdesk does their thing?

I'm pretty indifferent? If we merge it it will post in our team channel until it gets fixed by helpdesk. However, on the other hand, we'd need to open another PR to set it to always alert anyways.

.github/workflows/conformance.yml
```diff
@@ -109,12 +109,13 @@ jobs:
...
-      #     status: ${{ job.status }}
-      #     notify_when: 'failure'
-      #   env:
-      #     SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}
+      - name: Report Status
+        if: always()
+        uses: ravsamhq/notify-slack-action@v1
+        with:
+          status: ${{ job.status }}
+          mention_groups: '!here'
```
andrewstucki, line 117: Is this supposed to be an `@here`? Also, this could potentially get pretty noisy (even though we're not planning on running these often)... do we really want to `@` anyone?

.github/workflows/conformance.yml
```diff
@@ -109,12 +109,13 @@ jobs:
...
-      # - name: Report Status
-      #   if: always()
-      #   uses: ravsamhq/notify-slack-action@v1
-      #   with:
-      #     status: ${{ job.status }}
-      #     notify_when: 'failure'
-      #   env:
-      #     SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}
+      - name: Report Status
+        if: always()
```
nathancoleman, line 113: Thought on only notifying when the trigger was a scheduled run / when the branch is `main`?

I've had PRs where I ran conformance over and over as I tweek things, and I'm always looking directly at the PR status. Wouldn't wanna spam our channel in Slack.

sarahalsmiller: I'd agree with that, I'll get that added.

.github/workflows/conformance.yml
andrewstucki: I'd be fine with that so long as our conformance test runs aren't flaky 😅 -- I suppose we can always tune it down if it gets too noisy

nathancoleman (approved):

PR #147, opened 2022-04-12
Light refactor on new error types

sarahalsmiller:
-Change how errors are stored in ResolutionErrrors struct to minimize places where  new errors need to be added

### How I've tested this PR:
go test ./...

go test ./...

### Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/service/resolver.go
```diff
@@ -137,23 +105,23 @@ func (r *ResolutionErrors) Flatten() (ServiceResolutionErrorType, error) {
...
 	}
 
-	if len(r.consulErrors) != 0 {
-		return ConsulServiceResolutionErrorType, errors.New(r.String())
-	}
-
-	if len(r.refNotPermittedErrors) != 0 {
-		return RefNotPermittedErrorType, errors.New(r.String())
+	//otherwise assume that the first error type in the map is the error we found
+	for errType := range r.errors {
```
andrewstucki, line 114: so, Go map iteration order [isn't meant to be stable](https://go.dev/blog/maps#iteration-order):

> When iterating over a map with a range loop, the iteration order is not specified and is not guaranteed to be the same from one iteration to the next.

In practice it's actually [based off of hash keys and collision indexes](https://github.com/golang/go/blob/19309779ac5e2f5a2fd3cbb34421dafb2855ac21/src/runtime/map.go#L23-L31). So, if we're trying to do any sort of prioritization here, we'll need to specify an order via something like:

```go
// this array determines the priority of the error types returned
// when the resolution errors are flattened
var orderedResolutionErrorTypes = []ServiceResolutionErrorType{
  K8sServiceResolutionErrorType,
  ConsulServiceResolutionErrorType,
  RefNotPermittedErrorType,
  GenericResolutionErrorType,
}

...

func (r *ResolutionErrors) Flatten() (ServiceResolutionErrorType, error) {
  if r.Empty() {
    return NoResolutionErrorType, nil
  }

  for _, errorType := range orderedResolutionErrorTypes {
    if _, found := r.refNotPermittedErrors[errorType]; found {
      return errType, errors.New(r.String())
    }
  }

  // should not be reached
  return NoResolutionErrorType, nil
}
```

You can add the ordering array right at the top near the type specification just to make it so you don't have to track things down across the file.

internal/k8s/service/resolver.go
```diff
@@ -55,80 +55,48 @@ func NewRefNotPermittedError(inner string) ResolutionError {
 	return ResolutionError{inner, RefNotPermittedErrorType}
 }
 
+func getErrorTypePrefix(errType ServiceResolutionErrorType) string {
```
andrewstucki, line 58: if we wanted to make this more declarative too just to reduce the amount of tracking down helpers and switch/cases, you could move this into something like a `errorTypePrefixMap` that you could also just declare at the beginning of the file.

andrewstucki, line 58: If we wanted to make this more declarative too just to reduce the amount of tracking down helpers and switch/cases, you could move this into something like a `errorTypePrefixMap` that you could also just declare at the beginning of the file, but just a thought, no need for the change if you prefer the switch/case.

internal/k8s/service/resolver.go
```diff
@@ -137,23 +105,23 @@ func (r *ResolutionErrors) Flatten() (ServiceResolutionErrorType, error) {
...
 
-	if len(r.consulErrors) != 0 {
-		return ConsulServiceResolutionErrorType, errors.New(r.String())
-	}
-
-	if len(r.refNotPermittedErrors) != 0 {
-		return RefNotPermittedErrorType, errors.New(r.String())
+	//otherwise assume that the first error type in the map is the error we found
+	for errType := range r.errors {
+		fmt.Println(errType)
```
andrewstucki, line 115: whoops

andrewstucki (changes requested):
Overall, I like the simplification. There will be some issues around map iteration order that will need to get fixed, but generally good work!

nathancoleman (commented):
This looks great! The stray `fmt.Println` is the only thing holding back my approval

internal/k8s/service/resolver.go
```diff
@@ -55,80 +55,48 @@ func NewRefNotPermittedError(inner string) ResolutionError {
...
 func (r ResolutionError) Error() string {
 	return r.inner
 }
 
 type ResolutionErrors struct {
-	k8sErrors             []ResolutionError
-	consulErrors          []ResolutionError
-	genericErrors         []ResolutionError
-	refNotPermittedErrors []ResolutionError
+	errors map[ServiceResolutionErrorType][]ResolutionError
```
mikemorris, line 72: This might be pulling at a thread that unravels an entire sweater (and would require significantly more refactoring), but I'm curious if there's actually a reason to break up these errors by map keys, as opposed to something like directly using
```
    errrors []ResolutionError
```
instead of this entire struct, and adding fields like
```
type ResolutionError struct {
    origin: ConsulErrorType | K8sErrorType
    type: ServiceResolutionError | RefNotPermittedError
```

PR #148, opened 2022-04-15
Upgrade to go 1.17.9

andrewstucki:

This upgrades our build/distribution pipelines to use go 1.17.9 -- we were already actually testing using go 1.17.x. The major changes with the `go.mod` file come from running `go mod tidy` after specifying 1.17 in the file -- go 1.17 introduced a separate section in the file for indirect dependencies, which is what we now see.

Just so we know exactly what version of go we're shipping with, I pinned the patch version in the workflow as well.


.github/workflows/build.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       matrix:
         goos: ["linux"]
         goarch: ["arm", "arm64", "386", "amd64"]
-        go: ["1.16"]
+        go: ["1.17.9"]
```
mikemorris, line 66: Do we actually want to pin this to a patch version, or just 1.17 latest?
```suggestion
        go: ["1.17"]
```

.github/workflows/docker.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         goos: ["linux"]
         goarch: ["amd64"]
-        go: ["1.16"]
+        go: ["1.17.9"]
```
mikemorris, line 14:
```suggestion
        go: ["1.17"]
```

mikemorris (approved):

nathancoleman (approved):

.github/workflows/build.yml
andrewstucki: Going to just pin the patch for now, we can re-evaluate down the line if we so desire.

PR #152, opened 2022-04-18
Upgrade default envoy image to use 1.22 distroless

andrewstucki:

Upgrades the envoy we use everywhere to use [distroless releases](https://hub.docker.com/r/envoyproxy/envoy-distroless), which run without a shell. Not entirely sure if tests will pass initially as we may need to update our [entrypoint template](https://github.com/hashicorp/consul-api-gateway/blob/7af23d42414c6651d73e4ee941a866073b3c8c0d/internal/k8s/builder/gateway.go#L315) as it uses a shell script to inject Consul's CA certificate. We can probably fix this using environment variables directly on the deployment definition.

### DIFF
--- a/internal/k8s/builder/builder.go
+++ b/internal/k8s/builder/builder.go
@@ -26,7 +26,7 @@ func init() {
 }
 
 const (
-	defaultEnvoyImage     = "envoyproxy/envoy:v1.20-latest"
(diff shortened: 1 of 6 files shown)


nathancoleman:
This will work with #391

PR #155, opened 2022-04-21
Consume consul:1.12.0 for CI instead of 1.12.0-beta1

nathancoleman:
Consume consul.1.12.0 where we were previously consuming 1.12.0-beta1 for CI

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,9 @@ jobs:
       matrix:
         consul-version: 
         - 1.11.5
-        - 1.12.0-beta1
+        - 1.12.0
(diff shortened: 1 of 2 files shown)


andrewstucki (approved):
Do we need any corresponding changes in our conformance pipeline? Or is that mainly controlled by the helm chart defaults?

nathancoleman:
@andrewstucki the only change required there is the `consul-config.yaml` file in this PR. We won't even need that once consul 1.12.0 is consumed in consul-k8s

PR #156, opened 2022-04-21
http-route-controller: watch ReferencePolicy lifecycle

mikemorris:

Adds a `Watches` clause to the HTTPRouteReconciler to watch for changes to ReferencePolicy objects and trigger reconciliation and re-validation for any HTTPRoutes that may be affected by changes.

(Please don't view by commit, it's too messy to bother trying to interactive rebase cleanly 🙈)

### How I've tested this PR:
- [x] Added a unit test for `GatewayClient.GetHTTPRoutesInNamespace`
- [x] Added a unit test for `HTTPRouteReconciler.referencePolicyToRouteRequests`
- [ ] **TODO:** Add end-to-end tests for adding, updating and deleting a ReferencePolicy 

### Reference Documentation:
- https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.ReferencePolicySpec
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/controller#Controller
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/handler#EnqueueRequestsFromMapFunc
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/event#pkg-overview
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/client#InNamespace

For unimplemented GroupKind matching on ReferencePolicyFrom instead of assuming HTTPRoute, and BackendRef filtering on ReferencePolicyTo:
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/client#MatchingFieldsSelector
- https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Selector
- https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/controllers/http_route_controller.go
+++ b/internal/k8s/controllers/http_route_controller.go
@@ -3,7 +3,15 @@ package controllers
 import (
 	"context"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 	gateway "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/gatewayclient"
@@ -53,5 +61,143 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gateway.HTTPRoute{}).
+		Watches(
+			// TODO: how does this handle deletes?
+			&source.Kind{Type: &gateway.ReferencePolicy{}},
+			handler.EnqueueRequestsFromMapFunc(r.referencePolicyToRouteRequests),
+		).
 		Complete(gatewayclient.NewRequeueingMiddleware(r.Log, r))
+}
+
+func getReferencePolicyObjectsFrom(refPolicy gateway.ReferencePolicy) []client.Object {
+	matches := []client.Object{}
+
+	for _, from := range refPolicy.Spec.From {
+		matchLabels := map[string]string{
+			"kubernetes.io/metadata.kind":      string(from.Kind),
+			"kubernetes.io/metadata.namespace": string(from.Namespace),
+		}
+
+		if from.Group != "" {
+			matchLabels["kubernetes.io/metadata.group"] = string(from.Group)
+		} else {
+			// When empty, the Kubernetes core API group is inferred.
+			matchLabels["kubernetes.io/metadata.group"] = "core/v1"
+		}
+
+		selector := metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{{
+				Key:      "",
+				Operator: "In",
+				Values:   []string{},
+			}},
+			MatchLabels: matchLabels,
+		}
+
+		matches = append(matches, []client.Object{}...)
+	}
+
+	return matches
+}
+
+func getReferencePolicyObjectsTo(refPolicy gateway.ReferencePolicy) []client.Object {
+	matches := []client.Object{}
+
+	for _, to := range refPolicy.Spec.To {
+		selector := labels.NewSelector()
+
+		kindReq, _ := labels.NewRequirement(
+			"kubernetes.io/metadata.kind",
+			selection.In,
+			[]string{string(to.Kind)},
+		)
+
+		namespaceReq, _ := labels.NewRequirement(
+			"kubernetes.io/metadata.namesapce",
+			selection.In,
+			[]string{refPolicy.Namespace},
+		)
+
+		selector = selector.Add(*kindReq, *namespaceReq)
+
+		var groupReq *labels.Requirement
+		if to.Group != "" {
+			groupReq, _ = labels.NewRequirement(
+				"kubernetes.io/metadata.group",
+				selection.In,
+				[]string{string(to.Group)},
+			)
+		} else {
+			// When empty, the Kubernetes core API group is inferred.
+			groupReq, _ = labels.NewRequirement(
+				"kubernetes.io/metadata.group",
+				selection.In,
+				[]string{"core/v1"},
+			)
+		}
+
+		selector = selector.Add(*groupReq)
+
+		if to.Name != nil {
+			nameReq, _ := labels.NewRequirement(
+				"kubernetes.io/metadata.name",
+				selection.In,
+				[]string{string(*to.Name)},
+			)
+			selector = selector.Add(*nameReq)
+		}
+
+		// TODO: use selector
+		matches = append(matches, []client.Object{}...)
+	}
+
+	return matches
+}
+
+func GetRoutesAffectedByReferencePolicy(refPolicy gateway.ReferencePolicy) []gateway.HTTPRoute {
+	matches := []gateway.HTTPRoute{}
+	objectsTo := getReferencePolicyObjectsTo(refPolicy)
+
+	// FIXME: Only checking Routes selected by GetReferencePolicyObjectsFrom isn't
+	// enough, need to reconcile Routes which may have been allowed before but no
+	// longer are permitted
+	//
+	// GET Routes WHERE BackendRef is selected by GetReferencePolicyObjectsTo
+	for _, route := range getReferencePolicyObjectsFrom(refPolicy) {
+		// TODO: should this use reflection to handle xRoute types? seems expensive
+		for _, rules := range route.Spec.Rules {
+			for _, backendRef := range rules.BackendRefs {
+				for _, from := range objectsTo {
+					if backendRef.DeepEqual(from) {
+						matches = append(matches, backendRef)
+					}
+				}
+			}
+		}
+	}
+
+	return matches
+}
+
+func (r *HTTPRouteReconciler) referencePolicyToRouteRequests(object client.Object) []reconcile.Request {
+	r.Log.Info("event for ReferencePolicy", "object", object)
+
+	refPolicy := gateway.ReferencePolicy{}
+	routes := GetRoutesAffectedByReferencePolicy(refPolicy)
+	requests := []reconcile.Request{}
+
+	for _, route := range routes {
+		requests = append(requests, reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      route.Name,
+				Namespace: route.Namespace,
+			},
+		})
+	}
+
+	if len(requests) > 0 {
+		return requests
+	}
+
+	return nil
 }
--- a/internal/k8s/controllers/http_route_controller.go
+++ b/internal/k8s/controllers/http_route_controller.go
@@ -154,7 +154,6 @@ func getReferencePolicyObjectsTo(refPolicy gateway.ReferencePolicy) []client.Obj
 	return matches
 }
 
-// For
 func (r *HTTPRouteReconciler) getRoutesAffectedByReferencePolicy(refPolicy gateway.ReferencePolicy) []gateway.HTTPRoute {
 	matches := []gateway.HTTPRoute{}
 
--- a/internal/k8s/controllers/http_route_controller.go
+++ b/internal/k8s/controllers/http_route_controller.go
@@ -69,6 +69,7 @@ func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(gatewayclient.NewRequeueingMiddleware(r.Log, r))
 }
 
+// FIXME: create client.MatchingLabels here instead
 func getReferencePolicyObjectsFrom(refPolicy gateway.ReferencePolicy) []client.Object {
 	matches := []client.Object{}
 
@@ -100,6 +101,7 @@ func getReferencePolicyObjectsFrom(refPolicy gateway.ReferencePolicy) []client.O
 	return matches
 }
 
+// FIXME: create client.MatchingLabels here instead
 func getReferencePolicyObjectsTo(refPolicy gateway.ReferencePolicy) []client.Object {
 	matches := []client.Object{}
 
--- a/internal/k8s/controllers/http_route_controller.go
+++ b/internal/k8s/controllers/http_route_controller.go
@@ -3,7 +3,7 @@ package controllers
 import (
 	"context"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	// metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
@@ -85,14 +85,14 @@ func getReferencePolicyObjectsFrom(refPolicy gateway.ReferencePolicy) []client.O
 			matchLabels["kubernetes.io/metadata.group"] = "core/v1"
 		}
 
-		selector := metav1.LabelSelector{
-			MatchExpressions: []metav1.LabelSelectorRequirement{{
-				Key:      "",
-				Operator: "In",
-				Values:   []string{},
-			}},
-			MatchLabels: matchLabels,
-		}
+		// selector := metav1.LabelSelector{
+		// 	MatchExpressions: []metav1.LabelSelectorRequirement{{
+		// 		Key:      "",
+		// 		Operator: "In",
+		// 		Values:   []string{},
+		// 	}},
+		// 	MatchLabels: matchLabels,
+		// }
 
 		matches = append(matches, []client.Object{}...)
 	}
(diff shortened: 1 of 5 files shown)


internal/k8s/controllers/http_route_controller_test.go
```diff
@@ -75,3 +85,104 @@ func TestHTTPRoute(t *testing.T) {
 		})
 	}
 }
+
+func TestHTTPRouteWatchesReferencePolicy(t *testing.T) {
```
mikemorris, line 89:
```suggestion
func TestHTTPRouteReferencePolicyToRouteRequests(t *testing.T) {
```

internal/k8s/controllers/http_route_controller_test.go
```diff
@@ -75,3 +85,104 @@ func TestHTTPRoute(t *testing.T) {
...
+		ControllerName: mockControllerName,
+		Manager:        reconcilerMocks.NewMockReconcileManager(ctrl),
+	}
+
+	requests := controller.referencePolicyToRouteRequests(&refPolicy)
+
+	// FIXME: This should only request reconciliation for the one HTTPRoute in the
+	// allowed namespace
+	//
+	// require.Equal(t, 1, len(requests))
```
mikemorris, line 161:
```suggestion
```

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +60,126 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+// state in case a ReferencePolicy has been modified to revoke permission from a
+// namespace or to a service
+//
+// It may be possible to improve performance here by filtering Routes by
+// BackendRefs selectable by the To fields, but currently we just revalidate
+// all Routes allowed in the From Namespaces
+func (r *HTTPRouteReconciler) referencePolicyToRouteRequests(object client.Object) []reconcile.Request {
+	// TODO: Is there a safer way I could typecheck this with
+	// object.GetObjectKind() or something before casting?
+	refPolicy := *object.(*gateway.ReferencePolicy)
```
andrewstucki, line 83: this is how you'd do the safe typecast, but you may not need to given the fact that the `Watches` ensures the proper typing.

```suggestion
	refPolicy, ok := object.(*gateway.ReferencePolicy)
	if !ok {
		return nil
	}
```

Also, is there a reason for dereferencing after the cast?

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +60,126 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	return matches
+}
+
+func (r *HTTPRouteReconciler) getReferencePolicyObjectsFrom(refPolicy gateway.ReferencePolicy) []gateway.HTTPRoute {
+	matches := []gateway.HTTPRoute{}
+
+	for _, from := range refPolicy.Spec.From {
+		// TODO: search by from.Group and from.Kind instead of assuming HTTPRoute
+		routes, err := r.Client.GetHTTPRoutesInNamespace(context.TODO(), string(from.Namespace))
```
andrewstucki, line 169: you may want to pass the top-level context and store it on the reconciler struct so you can reference it here, just so we get clean cancelation of any outstanding requests during shutdown.

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +60,126 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if err != nil {
+			// TODO: is there a better way to handle this error?
+			return matches
```
andrewstucki, line 172: Since you can't really return an error here we should at least log it.

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +60,126 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// 	group := "core/v1"
+	// 	if to.Group != "" {
+	// 		group = string(to.Group)
+	// 	}
+
+	// 	toSelectors = append(toSelectors, groupKindToFieldSelector(schema.GroupKind{
+	// 		Group: group,
+	// 		Kind:  string(to.Kind),
+	// 	}))
+	// }
```
andrewstucki, line 116: This stuff can get deleted for now, right? Or are you intending to add logic back to filter down to a smaller subset of routes that need reconciliation?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -408,7 +408,7 @@ func TestHTTPMeshService(t *testing.T) {
 			serviceFive, err := e2e.DeployHTTPMeshService(ctx, cfg)
 			require.NoError(t, err)
 
-			namespace := e2e.Namespace(ctx)
+			namespace := e2e.Namespaces(ctx)[0]
```
andrewstucki, line 411: This change feels a bit dirty. If we want a specific cross-namespace test, I'd just create a new namespace directly in the test that calls for it.

andrewstucki (commented):
Some initial feedback

internal/k8s/controllers/http_route_controller.go
mikemorris: The deref is for the following call to `r.getRoutesAffectedByReferencePolicy(refPolicy)` (could deref at call site instead I guess)

andrewstucki: Pretty sure this is guaranteed never to be `nil` with how the controller runtime calls things (and if it was `nil`, this could potentially panic with a `nil` pointer dereference), but the only real use comes from iterating over `refPolicy.Spec.From` which shouldn't require the dereference. I'd just drop it and change the function signatures to take a pointer.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -964,6 +964,192 @@ func TestTCPMeshService(t *testing.T) {
...
+						Kind:  apigwv1alpha1.GatewayClassConfigKind,
+						Name:  configName,
+					},
+				},
+			}
+			err = resources.Create(ctx, gc)
+			require.NoError(t, err)
+
+			// FIXME: should this use a different port?
+			checkPort := e2e.HTTPFlattenedPort(ctx)
```
andrewstucki, line 1027: Yeah, so pretty sure that due to the way we bind/port-forward things in the `kind` setup that you'll have issues spinning up this gateway/gateways in other tests (i.e. you should see errors in the deployment that gets created due to not being able to bind to the node port and the pod shouldn't get scheduled). Generally in the `e2e` tests, **_due to port binding, you basically need to specify a new port for each gateway that you want to provision and issue requests through_**.

This is solvable with something like `MetalLB` allowing for dynamic allocation of ip addresses and dropping our port-forwarding in `kind` (and our use of `useHostPorts`), but I'm hesitant to add that because:

1) It makes the setup all the more complicated, and
2) On Windows/Mac `MetalLB` doesn't work out of the box due to offloading Docker to an external VM that doesn't actually share the same network as the host (you'd have to install something like https://github.com/chipmk/docker-mac-net-connect to run the e2e tests locally)

andrewstucki (changes requested):
So, the general scope of the commented out portions of your test look like more-or-less what I'd do (i.e. create the route in an invalid ref state, ensure status/routability check --> create the policy, re-check --> delete the policy re-check --> cleanup), but I think you're likely experiencing test failures due to re-using the same port as another test that binds a gateway to the same host port.

Have you tried running this locally with the commented out portions uncommented?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -964,6 +964,192 @@ func TestTCPMeshService(t *testing.T) {
...
+			port := gateway.PortNumber(serviceOne.Spec.Ports[0].Port)
+			route := &gateway.HTTPRoute{
+				ObjectMeta: meta.ObjectMeta{
+					Name:      routeName,
+					Namespace: routeNamespace,
+				},
+				Spec: gateway.HTTPRouteSpec{
+					CommonRouteSpec: gateway.CommonRouteSpec{
+						ParentRefs: []gateway.ParentRef{{
+							Name: gateway.ObjectName(gatewayName),
```
andrewstucki, line 1050: Dug into why this test is currently failing. This `parentRef` has a `Name`, but no `Namespace` value and the `HTTPRoute` is in a different namespace than the `Gateway` you're trying to bind it to. Because the controller can't find a managed `Gateway` object in the Route's namespace, it just skips the route since it doesn't think that it should do anything with it.

Think you just need to add a `Namespace` value here and the binding should work.

mikemorris: Thanks fro taking a look, that did the trick!

internal/k8s/controllers/http_route_controller.go
mikemorris: Hoping to uncomment this and enable the finer-grained filtering _if_ it would make sense to do so with regard to potential performance benefits (if the cost of iterating over all the backends of potentially-affected routes would be cheaper than just triggering a blanket reconciliation and re-validation on all of them).

internal/commands/server/k8s_e2e_test.go
mikemorris: Yea that makes sense, I was initially trying to put the services in a second namespaces and so much of that setup logic is abstracted away in the stack creation.

internal/commands/server/k8s_e2e_test.go
mikemorris: This feels like a thing we should scope out and take on at some point in the future if we can make the UX nice, but 👍 to just poking another hole for now.

internal/k8s/controllers/http_route_controller.go
andrewstucki: I think that generally this wouldn't be too bad in terms of performance since at least some of (if not all due to coming right after a cache fill via the `list` operation?) the triggered routes should hit the internal client cache for the controller runtime client.

I'd just drop the logic for now so long as it's working and maybe add in a comment to the function about potentially optimizing down the line to only enqueue the minimal subset of routes for re-reconciliation.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1018,48 +1228,47 @@ func tcpRouteStatusCheck(ctx context.Context, resources *resources.Resources, ga
...
 		}
+		return true
 	}
-	return false
 }
 
 func conditionAccepted(conditions []meta.Condition) bool {
-	for _, condition := range conditions {
-		if condition.Type == "Accepted" ||
-			condition.Status == "True" {
```
mikemorris, line 1035: Refactor uncovered that this should've been `&&`, so this check was erroneously passing on any condition with `Status: "True"`, and was being called in a few places where `conditionReady` should have been called instead.

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,125 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+// This is needed to reconcile any objects matched by  both current and prior
+// state in case a ReferencePolicy has been modified to revoke permission from a
+// namespace or to a service
+//
+// It may be possible to improve performance here by filtering Routes by
+// BackendRefs selectable by the To fields, but currently we just revalidate
+// all Routes allowed in the From Namespaces
+func (r *HTTPRouteReconciler) referencePolicyToRouteRequests(object client.Object) []reconcile.Request {
+	// TODO: Is there a safer way I could typecheck this with
+	// object.GetObjectKind() or something before casting?
```
mikemorris, line 83:
```suggestion
	// TODO: Is there a safer way I could typecheck this with
	// object.GetObjectKind() or something before casting?
```

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,123 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			&source.Kind{Type: &gateway.ReferencePolicy{}},
+			handler.EnqueueRequestsFromMapFunc(r.referencePolicyToRouteRequests),
+		).
 		Complete(gatewayclient.NewRequeueingMiddleware(r.Log, r))
 }
+
+// For UpdateEvents which contain both a new and old object, this transformation
+// function is run on both objects and both sets of Requests are enqueued.
+//
+// This is needed to reconcile any objects matched by  both current and prior
```
nathancoleman, line 74:
```suggestion
// This is needed to reconcile any objects matched by both current and prior
```

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,123 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gateway.HTTPRoute{}).
+		Watches(
+			&source.Kind{Type: &gateway.ReferencePolicy{}},
+			handler.EnqueueRequestsFromMapFunc(r.referencePolicyToRouteRequests),
+		).
```
nathancoleman, line 67: Do we need this same watch for `TCPRoutes`?

nathancoleman: I see https://github.com/hashicorp/consul-api-gateway/pull/156/files#diff-764889fcb4f05bae8ffc1396db1a0fc5f2c7d088ef9498edb3b546084d2fa936R166 and am wondering if we're intentionally only dealing with `HTTPRoutes` for this release or will need a followup PR before we release v0.2.0

mikemorris: I had been intending to limit the scope of this to `HTTPRoute` for the initial implementation - there are several TODOs throughout for refactoring/generalizing to other route types but I didn't want to bloat this scope initially.

nathancoleman: Do you know if we need to generalize ahead of the release, or is that a thing we can do for 0.3.0?

nathancoleman: Discussed in Slack that we want to handle for TCPRoutes as well

mikemorris: Discussed that we want to get this in for the same release as this logic for HTTPRoute, but planning the implement as a separate PR following this one.

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,73 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+// function is run on both objects and both sets of Requests are enqueued.
+//
+// This is needed to reconcile any objects matched by both current and prior
+// state in case a ReferencePolicy has been modified to revoke permission from a
+// namespace or to a service
+//
+// It may be possible to improve performance here by filtering Routes by
+// BackendRefs selectable by the To fields, but currently we just revalidate
+// all Routes allowed in the From Namespaces
+func (r *HTTPRouteReconciler) referencePolicyToRouteRequests(object client.Object) []reconcile.Request {
```
nathancoleman, line 81: So this will invalidate a route that was previously validate if a ReferencePolicy is deleted and the cross-namespace route is no longer permitted, right? We had talked about this in the last kube-sig meeting but I don't remember where exactly we landed.

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,73 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			return matches
+		}
+
+		matches = append(matches, routes...)
+	}
+
+	return matches
+}
+
+func groupKindToFieldSelector(gk schema.GroupKind) fields.Selector {
```
nathancoleman, line 129: Are we using this anywhere right now, or was it being used by the commented/removed code?

internal/k8s/controllers/http_route_controller.go
andrewstucki: the flow for this should be that the re-triggered route will wind up going through the whole resolution/binding shebang again, this time hitting the `RefNotPermitted` case:

https://github.com/hashicorp/consul-api-gateway/blob/df21465e8cb1932786a116052aff6deba233dae2/internal/k8s/reconciler/route.go#L209

internal/k8s/controllers/http_route_controller_test.go
```diff
@@ -75,3 +86,100 @@ func TestHTTPRoute(t *testing.T) {
...
+
+	builder := fake.
+		NewClientBuilder().
+		WithScheme(scheme)
+	if list != nil {
+		builder = builder.WithLists(list)
+	}
+	if len(objects) > 0 {
+		builder = builder.WithObjects(objects...)
+	}
```
nathancoleman, line 182: `objects` will be an empty list if nothing is passed in, so you don't need a check on `len`

```suggestion
	builder = builder.WithObjects(objects...)
```

internal/k8s/controllers/http_route_controller.go
nathancoleman: will wanna make sure this is documented somewhere

internal/k8s/controllers/http_route_controller_test.go
nathancoleman, line 182: `objects` will be an empty list if nothing is passed in, so you don't need a check on `len`
```suggestion
	builder := fake.
		NewClientBuilder().
		WithScheme(scheme).
		WithObjects(objects...)
	if list != nil {
		builder = builder.WithLists(list)
	}
```

nathancoleman (approved):
A few comments above but nothing to block merge IMO

internal/k8s/controllers/http_route_controller.go
```diff
@@ -53,5 +61,123 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	return matches
+}
+
+func (r *HTTPRouteReconciler) getReferencePolicyObjectsFrom(refPolicy *gateway.ReferencePolicy) []gateway.HTTPRoute {
+	matches := []gateway.HTTPRoute{}
+
+	for _, from := range refPolicy.Spec.From {
+		// TODO: search by from.Group and from.Kind instead of assuming HTTPRoute
+		routes, err := r.Client.GetHTTPRoutesInNamespace(r.Context, string(from.Namespace))
```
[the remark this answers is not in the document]
sarahalsmiller: After a few hours of digging through the dynamic and unstructured packages, I don't think it's strictly possible to search for multiple kinds in a single API call with the go-client (would love to be proven wrong here though). The most generic thing I can figure out how to do is potentially use a mapper to search for the routes that are available on the cluster and list each of those found group version kinds and apply our logic from there, but I'm not sure that even makes sense at the moment when we're only supporting two kinds.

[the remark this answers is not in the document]
nathancoleman: I'm thinking it's not worth more digging vs. just having `r.Client.GetTCPRoutesInNamespace( ... )`. Could wait and dig more if we wind up hitting rule of three

internal/k8s/controllers/http_route_controller_test.go
```diff
@@ -75,3 +86,100 @@ func TestHTTPRoute(t *testing.T) {
 		})
 	}
 }
+
+func TestHTTPRouteReferencePolicyToRouteRequests(t *testing.T) {
```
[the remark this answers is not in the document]
mikemorris: Done in #162

internal/k8s/gatewayclient/gatewayclient.go
```diff
@@ -250,6 +251,15 @@ func (g *gatewayClient) GetHTTPRoute(ctx context.Context, key types.NamespacedNa
...
 }
 
+// TODO: Make this generic over Group and Kind, returning []client.Object
+func (g *gatewayClient) GetHTTPRoutesInNamespace(ctx context.Context, ns string) ([]gateway.HTTPRoute, error) {
+	routeList := &gateway.HTTPRouteList{}
+	if err := g.Client.List(ctx, routeList, client.InNamespace(ns)); err != nil {
+		return []gateway.HTTPRoute{}, NewK8sError(err)
+	}
+	return routeList.Items, nil
+}
```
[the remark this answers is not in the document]
mikemorris: Done in #162

internal/k8s/controllers/http_route_controller.go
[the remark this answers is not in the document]
mikemorris: 👍 to revisiting this when we get to implementing [TLSRoute](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.TLSRoute) or [UDPRoute](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.UDPRoute) - something more generic _might_ be possible with `unstructured.Unstructured` struct and `scheme.Runtime` using reflecting to convert from `kind` strings to types as described in https://iximiuz.com/en/posts/kubernetes-api-go-types-and-common-machinery/ (and likely adding a RouteController interface), but definitely not straightforward.

Implemented as `GetTCPRoutesInNamespace` in #162

internal/k8s/controllers/http_route_controller.go
mikemorris: Yea, this is the big caveat I called out as needing review - with our current logic any BackendRef becoming unpermitted will cause the entire Route to be removed from the listener, likely disrupting traffic. The upstream conformance tests discussion seems to be leaning heavily towards accepting partially-invalid routes and just serving 503 responses for the invalid backends to handle this scenario more safely.

internal/k8s/controllers/http_route_controller.go
mikemorris: Nope, this was leftover from the `toSelector` bit I removed.

internal/k8s/controllers/http_route_controller_test.go
mikemorris: Implemented in #162 which moves this into a dedicated test helper package.

internal/k8s/controllers/http_route_controller.go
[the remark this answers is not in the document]
andrewstucki: Rather than doing things with strings and `unstructured`, if you really want you could just make a generic `List`/`ListInNamespace` function on the interface and pass in the object you want retrieved directly. The k8s client uses reflection for retrieving lists/objects anyway. It'd handle something like `client.ListInNamespace(ctx, &gateway.HTTPRouteList{}, namespace)`. Which is a lot closer to the vanilla k8s client `List` method. `TCPRoute` would just then be `client.ListInNamespace(ctx, &gateway.TCPRouteList{}, namespace)`. The function signature would just specify a [`runtimeclient.ObjectList`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#ObjectList).

The main two reasons you'll still want to wrap the k8s client call, even if it's a really thin wrapper over `List` are for:

1. `K8sError` wrapping for our retry middleware
2. Easy mocking for error injection for tests

I don't necessarily have much of a preference but think the per-type signatures are kind of nice in that you can just having the initialized structures passed back to you. Per-type also follow more of the pattern throughout most of the `gatewayclient` interface. But either way is fine with me. If you _do_ change it though, I'd do that in a follow-up PR at this point, no need to re-block.

andrewstucki (approved):

internal/k8s/controllers/http_route_controller.go
nathancoleman: 🤦‍♂️ I see that called out in the PR body now. Thanks!

Seems like we're getting a little more pressure on handling our own xDS every day lately

PR #158, opened 2022-04-22
OpenAPI generation scaffolding

sarahalsmiller:
-Add scaffolding to go generate to generate API boilerplate from openAPI spec

### How I've tested this PR:
- running go generate ./...
- tests pass

- tests pass

### Checklist:
- [ X] Tests added
- [ ] CHANGELOG entry added 


go.mod
```diff
@@ -47,13 +48,18 @@ require (
 	github.com/containerd/containerd v1.5.7 // indirect
 	github.com/containerd/continuity v0.2.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/deepmap/oapi-codegen v1.10.1 // indirect
```
andrewstucki, line 51: So, this shouldn't get added here since nothing in the generated code relies on it, think we can revert most of the `go.mod` changes except for the stuff that pulls in `github.com/go-chi/chi/v5` in this commit. Everything else seems like it's just part of the global `go install` process which shouldn't actually affect this package's modules.

.github/workflows/ci.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       run: |
         echo "Checking that code generation is up-to-date"
         go install github.com/golang/mock/mockgen@v1.6.0
+        go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen
```
andrewstucki, line 63: 👍

nathancoleman (changes requested):

internal/api/schemas/v1.yaml
```diff
@@ -0,0 +1,47 @@
...
+  version: 1.0.0
+  title: Consul API Gateway
+  description: Consul API Gateway Management API
+  contact:
+    name: Consul API Gateway Team
+    email: consul-api-gateway@hashicorp.com
+    url: http://hashicorp.com
+  license:
+    name: Mozilla Public License 2.0
+    url: https://www.mozilla.org/en-US/MPL/2.0/
```
nathancoleman, line 12: I see that nomad-openapi is linking to the license in the repo. Wondering if we should do the same

https://github.com/hashicorp/nomad-openapi/blob/ccfbbe2cdd4a2e4d65db0204f9d3a5b51215d2b3/v1/openapi.yaml#L5658

nathancoleman (approved):

internal/api/schemas/v1.yaml
sarahalsmiller: I think it's effectively saying the same thing since the LICENSE.md is just a copy of MPL 2, but went ahead and updated to point to our repo.

PR #159, opened 2022-04-22
Add middleware to authenticate consul token

sarahalsmiller:
- add middleware to authenticate consul token to generated API

### How I've tested this PR:
-go test

- tests pass

### Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 


internal/api/middleware/oapi.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			// validate request
+			if statusCode, err := validateRequest(r, router, options); err != nil {
+				handler(w, statusCode, err.Error())
```
andrewstucki, line 43: This is basically the one line that differs from the [original upstream](https://github.com/deepmap/oapi-codegen/blob/9dc8b8d293a991614ea12447bd6507bfadf38304/pkg/chi-middleware/oapi_validate.go#L42). The function signatures also have `handler ErrorHandler` passed in to them to make it so that this validation middleware can respond with our standard error handling helper (the `sendError` call that makes proper JSON messages) rather than returning a plain text error. I'll look into extending the `Options` values and submitting a PR upstream so we can get rid of this, but until then, don't think it's a big deal to just lift this here.

andrewstucki: BTW, just opened [this upstream PR](https://github.com/deepmap/oapi-codegen/pull/557), so if that gets merged any time soon we can just drop this additional middleware and just use the upstream library code.

PR #160, opened 2022-04-25
Store Refactor - Part 1

andrewstucki:

Pretty much all of this is code taken from https://github.com/hashicorp/consul-api-gateway/pull/137 -- but I tried (somewhat) to make the code a bit more targeted -- though that's pretty difficult with some of these changes. Things introduced:

1. `factory` for doing the initialization of routes/gateways (and eventual serialization/deserialization)
2. `state` structures for trying to shore up some of the memoized status of the routes/gateways
3. `deployer` to remove much of the logic around ensuring deployment outside of the `Gateway` itself
4. dropping of unused UDP/TLS routes

Things that this _**does not**_ include that will come in subsequent PRs

1. Removal of binding and validation logic into `binder` and `validator` structures
2. Removal of listeners
3. Modification and concretization of the `Store` interface
4. Clean up of state setting and refactoring of tests to be more table-driven

### How I've tested this PR:

Unit and e2e tests pass, running conformance in this PR.


internal/k8s/reconciler/errors/generator.go
```diff
@@ -0,0 +1,183 @@
+//go:build generate
+// +build generate
+
+package main
```
andrewstucki, line 4: This file is basically half of the generator routines, I split up the errors and status generators into their own subpackages to try and begin to limit the scope of the top-level `reconciler` package.

internal/k8s/reconciler/gateway.go
```diff
@@ -48,40 +46,8 @@ type K8sGatewayConfig struct {
...
-		serviceBuilder:    service,
-		consulNamespace:   config.ConsulNamespace,
-		logger:            gatewayLogger,
-		client:            config.Client,
-		gateway:           gateway,
-		listeners:         listeners,
-	}
-}
-
 func (g *K8sGateway) Validate(ctx context.Context) error {
```
andrewstucki, line 49: Most of the logic from here down will eventually get moved into a dedicated `GatewayValidator` in a subsequent PR.

internal/k8s/reconciler/listener.go
```diff
@@ -51,7 +33,7 @@ type K8sListener struct {
 	listener        gw.Listener
 	client          gatewayclient.Client
 
-	status         ListenerStatus
+	status         *status.ListenerStatus
```
andrewstucki, line 36: For now, just passing a pointer to the status since all of the validation logic is contained here. Pretty much this entire struct is going to go away and we'll just set status values directly on `ListenerState` inside of the validator.

internal/k8s/reconciler/utils.go
```diff
@@ -204,18 +203,6 @@ func toNamespaceSet(name string, labels map[string]string) klabels.Labels {
 	return klabels.Set(ret)
 }
 
-func sortParents(parents []gw.RouteParentStatus) []gw.RouteParentStatus {
```
andrewstucki, line 207: Since pretty much all of these methods deal with setting status values on the route status, I moved them into a `RouteStatuses` struct that just wraps the underlying map

andrewstucki (commented):
Some initial comments

internal/k8s/reconciler/factory.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (f *Factory) NewGateway(config NewGatewayConfig) *K8sGateway {
+	gatewayState := config.State
+	if gatewayState == nil {
+		gatewayState = state.InitialGatewayState(config.Gateway)
+	}
+	return f.initializeGateway(&K8sGateway{
+		Gateway:         config.Gateway,
+		GatewayState:    gatewayState,
```
nickethier, line 55: Any reason `GatewayState` should be exported on K8sGateway? That seems like something that should be private

internal/k8s/reconciler/state/gateway.go
```diff
@@ -0,0 +1,104 @@
...
+		state.Listeners = append(state.Listeners, &ListenerState{
+			Name:     listener.Name,
+			Protocol: listener.Protocol,
+			Routes:   make(map[string]core.ResolvedRoute),
+		})
+	}
+	return state
+}
+
+func (g *GatewayState) GetStatus(gateway *gw.Gateway) gw.GatewayStatus {
```
nickethier, line 38: Don't forget to add a few tests in future PRs. This seems like a critical bit to ensure correctness of.

internal/k8s/reconciler/state/gateway.go
```diff
@@ -0,0 +1,104 @@
...
+		listenerStatuses = gateway.Status.Listeners
+	}
+	if status.ConditionsEqual(conditions, gateway.Status.Conditions) {
+		conditions = gateway.Status.Conditions
+	}
+
+	ipType := gw.IPAddressType
+	addresses := make([]gw.GatewayAddress, 0, len(g.Addresses))
+	for _, address := range g.Addresses {
+		addresses = append(addresses, gw.GatewayAddress{
```
nickethier, line 57: nit: if you're going to create a slice of length `len(g.Addresses)` might as well skip the append and just index off the for loop

nickethier (approved):
Code changes look good to me and I like the direction its going. Happy to be a sounding board for ideas on how to refactor binding/validation and the memory store.

internal/k8s/reconciler/factory.go
andrewstucki: So, the state is actually going to require serialization alongside the actual gateway definitions once we can do the full serializing. Basically a "stored" gateway is the definition + its bound state, so just to keep with common Go conventions, made anything that's "serializable" exported.

internal/k8s/reconciler/state/gateway.go
andrewstucki: 👍 the testing code for this is going to get a refactor along with moving status setting out of `gateway.go` and into a set of dedicated validators.

internal/k8s/reconciler/deployer.go
```diff
@@ -0,0 +1,121 @@
...
+				data, err := json.MarshalIndent(mutated, "", "  ")
+				if err == nil {
+					d.logger.Trace("created or updated gateway service", "service", string(data))
+				}
+			}
+		}
+	}
+
+	return nil
+}
```
nathancoleman, line 106: Similarly here

```suggestion
func (d *GatewayDeployer) ensureService(ctx context.Context, config apigwv1alpha1.GatewayClassConfig, gateway *gw.Gateway) error {
	service := d.Service(config, gateway)
	if service == nil {
	    return nil
	}
	
	mutated := service.DeepCopy()
	updated, err := d.client.CreateOrUpdateService(ctx, mutated, func() error {
		mutated = apigwv1alpha1.MergeService(service, mutated)
		return d.client.SetControllerOwnership(gateway, mutated)
	})
	if err != nil {
		return fmt.Errorf("failed to create or update gateway service: %w", err)
	}
	
	if updated && d.logger.IsTrace() {
		data, err := json.MarshalIndent(mutated, "", "  ")
		if err == nil {
			d.logger.Trace("created or updated gateway service", "service", string(data))
		}
	}
	
	return nil
}
```

internal/k8s/reconciler/deployer.go
```diff
@@ -0,0 +1,121 @@
...
+	if serviceAccount := config.ServiceAccountFor(gateway); serviceAccount != nil {
+		if err := d.client.EnsureServiceAccount(ctx, gateway, serviceAccount); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *GatewayDeployer) ensureDeployment(ctx context.Context, namespace string, config apigwv1alpha1.GatewayClassConfig, gateway *gw.Gateway) error {
+	deployment := d.Deployment(namespace, config, gateway)
```
nathancoleman, line 68: Should we be doing the same nil check on `deployment` here that we do for `service` in the next func?

internal/k8s/reconciler/deployer.go
```diff
@@ -0,0 +1,121 @@
...
+	return nil
+}
+
+func (d *GatewayDeployer) Deployment(namespace string, config apigwv1alpha1.GatewayClassConfig, gateway *gw.Gateway) *apps.Deployment {
+	deploymentBuilder := builder.NewGatewayDeployment(gateway)
+	deploymentBuilder.WithSDS(d.sdsHost, d.sdsPort)
+	deploymentBuilder.WithClassConfig(config)
+	deploymentBuilder.WithConsulCA(d.consulCA)
+	deploymentBuilder.WithConsulGatewayNamespace(namespace)
+	return deploymentBuilder.Build()
```
nathancoleman, line 114: #nit This feels odd for consuming a builder pattern. I'm used to seeing

```suggestion
	return builder.NewGatewayDeployment(gateway).
	    WithSDS(d.sdsHost, d.sdsPort).
	    WithClassConfig(config).
	    WithConsulCA(d.consulCA).
            WithConsulGatewayNamespace(namespace).
            Build()
```

internal/k8s/reconciler/errors/generator.go
nathancoleman, line 4: it feels odd that we have an executable buried so deep in our package structure. Can you add a docstring explaining that this does not run as part of our application?

I'm assuming it's located here for access reasons?

nathancoleman (changes requested):
Nice work! I'm about halfway through but figured I'd go ahead and share what I have so far.

internal/store/memory/store.go
```diff
@@ -177,56 +177,46 @@ func (s *Store) DeleteRoute(ctx context.Context, id string) error {
...
 
 	id := gateway.ID()
 
 	current, found := s.gateways[id]
+	var currentGW store.Gateway
+	if found {
+		currentGW = current.Gateway
+	}
 
-	if !current.ShouldUpdate(gateway) {
```
nathancoleman, line 222: We previously assumed `current` was non-nil. Why do we now assume it might be nil?

internal/k8s/service/resolver.go
```diff
@@ -169,31 +169,28 @@ func (r *ResolvedReference) Item() client.Object {
 }
 
 type BackendResolver interface {
-	Resolve(ctx context.Context, ref gw.BackendObjectReference) (*ResolvedReference, error)
+	Resolve(ctx context.Context, namespace string, ref gw.BackendObjectReference) (*ResolvedReference, error)
```
nathancoleman, line 172: Assuming this change is to have the resolver be more of a longstanding global that isn't tied to a single namespace like it was before?

internal/k8s/reconciler/errors/generator.go
andrewstucki: Sure, I can go ahead and add a more full-fledged comment apart from the:

```go
/*
This file generates boilerplate error structures
*/
```

that's a couple lines below this. Really though, since this is just for code generation, putting it anywhere outside of the subpackage that it's generating code for seems odd to me. The nice thing about having the build tags and the `go:generate` directly in this file is that it acts as any other typical `go generate` target and essentially just executes itself.

internal/k8s/reconciler/utils.go
```diff
@@ -278,55 +253,6 @@ func listenerStatusEqual(a, b gw.ListenerStatus) bool {
 	return conditionsEqual(a.Conditions, b.Conditions)
 }
 
-func listenerStatusesEqual(a, b []gw.ListenerStatus) bool {
```
nathancoleman, line 281: For future reviewers, these were all moved to `equality.go`

internal/k8s/reconciler/converters/http.go
```diff
@@ -1,40 +1,61 @@
...
+	}
+}
+
+func (c *HTTPRouteConverter) Convert() core.ResolvedRoute {
 	hostnames := []string{}
-	for _, hostname := range route.Spec.Hostnames {
+	for _, hostname := range c.route.Spec.Hostnames {
 		hostnames = append(hostnames, string(hostname))
 	}
 	if len(hostnames) == 0 {
```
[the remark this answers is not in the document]
andrewstucki: So, I'd prefer not to expand the scope of this too much by cleaning up/adding comments on already existing code, I think maybe I can document all the places where we can add some comments, but the refactor is less about filling in existing comments for clarity at the moment, and more for enabling functional changes, so I see something like this as a bit out of scope for the time being. I'm happy to do another PR though with the comment after we get all of this stuff merged.

internal/k8s/reconciler/converters/tcp.go
```diff
@@ -1,29 +1,49 @@
...
 	"github.com/hashicorp/consul-api-gateway/internal/core"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/reconciler/state"
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/service"
-	"k8s.io/apimachinery/pkg/types"
 	gw "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
-func TCPRouteID(namespacedName types.NamespacedName) string {
-	return "tcp-" + namespacedName.String()
+type TCPRouteConverter struct {
```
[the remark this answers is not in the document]
andrewstucki: It is actually converting from a `TCPRoute` (wrapped in a `K8sRoute`) to something else, the bound/"resolved view" of a route that's actually used to generate the consul synchronization code. The reason the conversion to this intermediate state needs to happen is to keep the synchronization code decoupled from the actual concrete implementations of Routes/Gateways -- we can model them however we want so long as we know how to convert them to the intermediate representation.

I can go ahead and add that to a comment, but this is really just shoring up the `convert*` functions that were previously there and uncommented into a dedicated "service" structure.

internal/k8s/reconciler/deployer.go
andrewstucki: Agree, and I'm fine with changing a bit of the usage of the builder pattern down the line, but don't quite consider it in scope right now for this pass.

internal/k8s/reconciler/converters/http.go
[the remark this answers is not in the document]
nathancoleman: That works 👍

internal/k8s/reconciler/deployer.go
andrewstucki: nope, a gateway will always have a deployment associated with it, it won't have a service unless you specifically specify a service type to use for exposing the gateway in the corresponding gatewayclassconfig

nathancoleman (changes requested):
Plugging away some more after meetings. Here's the next batch. Nearing the end!

internal/k8s/reconciler/status/route_status.go
```diff
@@ -0,0 +1,163 @@
...
+				Name: "other",
+			},
+			ControllerName: "other",
+		}},
+	}
+
+	statuses.Bound(refID)
+	merged := statuses.mergedStatus(status, "expected", 0).Parents
+	require.Len(t, merged, 3)
+	require.Equal(t, "expected", string(merged[0].ParentRef.Name))
```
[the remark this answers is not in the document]
nathancoleman: This might be one of those areas where you're just moving pre-existing code. No required change here

internal/k8s/reconciler/gateway.go
```diff
@@ -182,13 +152,13 @@ func (g *K8sGateway) assignGatewayIPFromServiceIngress(ctx context.Context, serv
...
-		g.status.Scheduled.NotReconciled = errors.New("service not found")
+		g.GatewayState.Status.Scheduled.NotReconciled = errors.New("service not found")
 		return nil
 	}
 
 	for _, ingress := range updated.Status.LoadBalancer.Ingress {
-		g.serviceReady = true
-		g.addresses = append(g.addresses, ingress.IP)
+		g.GatewayState.ServiceReady = true
+		g.GatewayState.Addresses = append(g.GatewayState.Addresses, ingress.IP)
```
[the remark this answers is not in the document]
andrewstucki: yep, that's the merge conflict atm 😄

internal/k8s/reconciler/gateway.go
```diff
@@ -96,21 +62,24 @@ func (g *K8sGateway) Validate(ctx context.Context) error {
...
+		}
 	}
 
 	return nil
 }
 
 type mergedListener struct {
 	port      gw.PortNumber
-	listeners []*K8sListener
+	listeners []int
```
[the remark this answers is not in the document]
andrewstucki: The incentive is to fully remove all semblance of `Listener` structures from our code, as happens in https://github.com/hashicorp/consul-api-gateway/pull/163 and https://github.com/hashicorp/consul-api-gateway/pull/165 -- the reason being is that my decision to pull out listeners into their own abstraction initially was a poor one due to them being configured directly as part of (and hence stored as part of) an individual Gateway's configuration.

Instead the movement is to just map part of the listener's configuration on the gateway spec to a corresponding area of the Gateway's state (that maintains the bound routes, etc.) -- as a result we just use indices to reference which listener we're talking about.

internal/k8s/reconciler/factory.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+}
+
+func (f *Factory) initializeGateway(gateway *K8sGateway) *K8sGateway {
+	logger := f.logger.Named("gateway").With("name", gateway.Name, "namespace", gateway.Namespace)
+	gateway.logger = logger
+	gateway.client = f.client
+	gateway.deployer = f.deployer
+
+	// get rid of this
```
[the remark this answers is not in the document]
andrewstucki: The `Factory` structure really existed to handle adding back in the minimal amount of non-serializable fields to our gateways/routes. The refactor, by the time you get to #165 is complete enough where I was actually fully able to remove this structure, so I'd ideally not have to go back and clean this up too much.

Initially I was trying to chunk a lot of this work and set things up for the piecemeal removal of things such as the Listener structures (as mentioned), this code acted as an intermediate state to allow me to keep the Listener code as-is until the refactor was fully done.

internal/k8s/reconciler/state/gateway.go
```diff
@@ -0,0 +1,104 @@
...
+	"errors"
+
+	"github.com/hashicorp/consul-api-gateway/internal/core"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/reconciler/common"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/reconciler/status"
+	gw "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// GatewayState holds ephemeral state for gateways
+type GatewayState struct {
```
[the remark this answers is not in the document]
andrewstucki: not until https://github.com/hashicorp/consul-api-gateway/pull/165 which adds in some of the serialization logic to some of the needed structures (like the `*Status` stuff).

internal/store/memory/store.go
andrewstucki: We hadn't assumed it was non-nil before, the logic was just hidden inside of the `ShouldUpdate` function on the `K8sGateway` structure:

https://github.com/hashicorp/consul-api-gateway/blob/564ad2b0114407168dfd0e9066bd463fd61a9efa/internal/k8s/reconciler/gateway.go#L352-L359

internal/k8s/service/resolver.go
andrewstucki: yeah, some of the small changes like this were to make it so that these service structures were usable without having to re-initialize them -- so they are no longer scoped to resolving a single route's backends (via its namespace), but instead can be initialized once and called for any route that needs to get resolved.

internal/k8s/reconciler/gateway.go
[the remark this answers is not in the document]
nathancoleman: I like this. Coming in knowing nothing a couple of months ago, I started with the K8s Gateway API and this would have made more sense

internal/k8s/reconciler/route.go
```diff
@@ -78,10 +62,6 @@ func (r *K8sRoute) ID() string {
 		return HTTPRouteID(utils.NamespacedName(r.Route))
 	case *gw.TCPRoute:
 		return TCPRouteID(utils.NamespacedName(r.Route))
-	case *gw.UDPRoute:
```
nathancoleman, line 81: Feels good to drop this mental overhead

nathancoleman (commented):
LGTM once the merge conflicts are resolved. Nice work! I appreciated the comments you dropped throughout when you opened the code review 💬

internal/k8s/reconciler/gateway.go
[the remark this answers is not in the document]
mikemorris: This feels like something we actually may want to think twice about - https://github.com/kubernetes-sigs/gateway-api/discussions/1248 discusses a proposal for [optionally] promoting Listeners to a full object, which could more closely map to our current implementation.

nathancoleman:
See list of referencing PRs above for the merging of the changes in this PR

PR #161, opened 2022-04-25
Specify explicit type config for Envoy bootstrap filters

nathancoleman:
Consul-K8s moved to Envoy 1.22 which includes a [minor behavior change](https://www.envoyproxy.io/docs/envoy/v1.22.0/version_history/current#minor-behavior-changes) that breaks due to our current bootstrap config:
> config: type URL is used to lookup extensions regardless of the name field. This may cause problems for empty filter configurations or mis-matched protobuf as the typed configurations. This behavioral change can be temporarily reverted by setting runtime guard envoy.reloadable_features.no_extension_lookup_by_name to false.

### How I've tested this PR:
🤖 conformance tests pass where they fail before

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/envoy/manager.go
+++ b/internal/envoy/manager.go
@@ -263,7 +263,10 @@ const bootstrapJSONTemplate = `{
                   },
                   "http_filters": [
                     {
-                      "name": "envoy.filters.http.router"
+                      "name": "envoy.filters.http.router",
+                      "typedConfig": {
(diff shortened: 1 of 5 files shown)


andrewstucki (approved):
LGTM on the actual change, pretty sure unit tests are going to fail due to the golden files being out of date though, so think you'll prob need to run `GENERATE=true go test` in  `internal/envoy` before merging.

PR #163, opened 2022-04-26
Store Refactor - Part 2

andrewstucki:

This is the second part of breaking up https://github.com/hashicorp/consul-api-gateway/pull/137. The _vast_ majority of the +- on the diff come from moving code around into dedicated service structures, like `binder`/`validator`. The things in this PR:

1. Add `Route`/`Gateway` `validator` structures to set the initial state for route binding
2. Extract the logic for binding into a dedicated `binder`
3. Remove all notions of "listeners" outside of `GatewayState`.
4. Shore up the `store` interface to reflect closer to the desired `backend` interface shape
5. Slightly modify the sync adapter interface to let the store know when it synced v. no-op'd so we can get rid of the additional memory listener wrappers

Things left to do:

1. Concretize the `store` implementation and introduce the `backend` interface.
2. Add serialization/deserialization via the new `factory`
3. Remove most of the remaining logic from `Route` into the binder/validators
4. Further clean up some of the unit tests.


internal/envoy/middleware.go
```diff
@@ -49,30 +50,20 @@ func GatewayFromContext(ctx context.Context) core.GatewayID {
 	return value.(core.GatewayID)
 }
 
-// GatewaySecretRegistry is used as the authority for determining what gateways the SDS server
-// should actually respond to because they're managed by consul-api-gateway
-type GatewaySecretRegistry interface {
```
andrewstucki, line 54: This abstraction gets dropped because we just use the store directly and check via the gateway itself whether or not it references the given secrets

internal/k8s/reconciler/gateway.go
```diff
@@ -46,304 +43,85 @@ type K8sGatewayConfig struct {
...
 		return nil
 	}
 
-	if pod.Status.HostIP != "" {
-		g.GatewayState.ServiceReady = true
-		g.GatewayState.Addresses = append(g.GatewayState.Addresses, pod.Status.HostIP)
-	}
-
-	return nil
+	return NewBinder(g.client, g.Gateway, g.GatewayState).Bind(ctx, k8sRoute)
```
andrewstucki, line 68: The Binder itself gets fully reworked in the 3rd PR to be re-usable across gateways/routes, but this is the interim state for now -- for some context, the movement towards extracting out all of the more primitive structures as arguments like the `*gw.Gateway` and `*state.GatewayState` was to be explicit about what was needed in what part of the binding/validation/resolution process just so I could more easily drop the fields once different parts of the refactor were complete.

internal/store/interfaces.go
```diff
@@ -32,39 +22,10 @@ type StatusTrackingGateway interface {
...
-// Listener describes the basic methods of a gateway
-// listener.
-type Listener interface {
-	ID() string
-	CanBind(ctx context.Context, route Route) (bool, error)
-	Config() ListenerConfig
-	IsValid() bool
+	Bind(ctx context.Context, route Route) []string
+	Remove(ctx context.Context, id string) error
+	Resolve() core.ResolvedGateway
```
andrewstucki, line 27: So, rather than calling `CanBind` on a listener, storing that result externally somewhere and then invoking `Resolve` on the route with the corresponding listener that could be bound. The new thing is basically calling `Bind` on the gateway with the route, storing that internally in the gateway state, and then just invoking `Resolve` directly on the gateway to grab the fully resolved state of the gateway that we need to sync over.

When the Binder itself gets modified a bit in the next PR, the `Bind` call gets fully extracted to a `Binder` interface to keep the `Gateway` a bit lighter.

PR #165, opened 2022-04-27
Store Refactor - Part 3

andrewstucki:

This is the third part in a series of refactors to get our storage into a place where we're able to handle serializable data. The things in this PR:

- Made a concrete store implementation
- Broke out a `Backend` interface and modified the memory store to use it
- Added some interfaces for:
1. status updating (to get rid of the overloaded Gateway/Store interfaces)
2. marshaling/umarshaling Routes and Gateways
3. handling the binding logic
- In the process, simplified the usage of `binder` so it doesn't need to be initialized on a per-gateway basis
- Finished stripping out all of the logic in K8sRoutes/Gateways such that there's no longer any need for a factory to initialize non-serializable structures
- Wrote some custom marshalers for otherwise non-serializable data

Things left to do:

- Further clean up some of the code and unit tests.
- Light final pass and shoring up any comment doc strings.

### How I've tested this PR:
Unit and e2e tests pass, running conformance tests here.


internal/store/interfaces.go
```diff
@@ -8,37 +8,11 @@ import (
...
-	Gateway
-
-	TrackSync(ctx context.Context, sync func() (bool, error)) error
-}
-
 // Gateway describes a gateway.
 type Gateway interface {
 	ID() core.GatewayID
-	Bind(ctx context.Context, route Route) []string
-	Remove(ctx context.Context, id string) error
```
andrewstucki, line 26: These are now the responsibility of the `Binder` interface with `Bind`/`Unbind` methods. `Unbind` also incorporates the logic of the `OnGatewayRemoved` method below.

internal/store/interfaces.go
```diff
@@ -8,37 +8,11 @@ import (
...
 
-// StatusTrackingGateway is an optional extension
-// of Gateway. If supported by a Store, when
-// a Gateway is synced to an external location,
-// its corresponding callbacks should
-// be called.
-type StatusTrackingGateway interface {
-	Gateway
-
-	TrackSync(ctx context.Context, sync func() (bool, error)) error
```
andrewstucki, line 19: This and `SyncStatus` get wrapped into a `StatusUpdater` interface that's optional to set for the concrete store implementation -- that way we can drop the overloaded `StatusTrackingGateway` and `StatusTrackingRoute`

internal/store/interfaces.go
```diff
@@ -8,37 +8,11 @@ import (
...
-}
-
 // Gateway describes a gateway.
 type Gateway interface {
 	ID() core.GatewayID
-	Bind(ctx context.Context, route Route) []string
-	Remove(ctx context.Context, id string) error
 	Resolve() core.ResolvedGateway
-	CanFetchSecrets(ctx context.Context, secrets []string) (bool, error)
```
andrewstucki, line 28: Removed the `Context`/`error` since really the secret checks are a property of the `Gateway` itself. This also gives us an interface that reflects a bit more of what we want -- just a bag of structured data rather than something that's potentially holding on to things like kubernetes/consul clients it can use for resolving extra bits.

internal/store/interfaces.go
```diff
@@ -47,12 +21,52 @@ type Route interface {
...
-// Store is used for persisting and querying gateways and routes
+type GatewayRecord struct {
+	ID   core.GatewayID
+	Data []byte
+}
+
+type RouteRecord struct {
+	ID   string
+	Data []byte
+}
```
andrewstucki, line 32: These are introduced so we have a simple way of doing bulk "upsert" operations. Since a route/gateway update can potentially affect a bunch of other routes/gateways, rather than updating them all individually, the idea is to pass off to the backend a list of records that need to be updated with:

1. their identifier (most likely used to construct some sort of key to their storage location), and
2. the marshaled bytes of the data to store

internal/store/interfaces.go
```diff
@@ -47,12 +21,52 @@ type Route interface {
...
+	Bind(ctx context.Context, gateway Gateway, route Route) (bool, error)
+	Unbind(ctx context.Context, gateway Gateway, route Route) (bool, error)
+}
+
+type StatusUpdater interface {
+	UpdateGatewayStatusOnSync(ctx context.Context, gateway Gateway, sync func() (bool, error)) error
+	UpdateRouteStatus(ctx context.Context, route Route) error
+}
+
 type Store interface {
```
andrewstucki, line 63: This is really only used for mocking/testing -- other than mocks, really only the concrete `*store` impllementation should be used.

andrewstucki (commented):
Some initial comments

nathancoleman:
See the linked PRs above which updated this refactor and merged it in small, incremental PRs

PR #167, opened 2022-04-27
ci: bump Go version to 1.18

mikemorris:

- Bump `GO_VERSION` for CI to 1.18
- Bump the `reflect2` dependency to avoid a segfault

### How I've tested this PR:

Running `go test ./...` in `internal/k8s/builder` previously failed with a segfault when running with Go 1.18 locally due to unsafe code used by the `reflect2` package (an indirect dependency from Kubernetes core). With this change tests now pass locally, and should be able to pass in CI as well.

- Confirm that bumping the version in `go.mod` is not necessary as we aren't actually using any features like generic that require Go 1.18 yet, just fixing compatibility.
- Did I miss anywhere this needs to be bumped, e.g. for release pipelines?

- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/167.txt
+++ b/.changelog/167.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+go: build with Go 1.18
+```
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: ["main", "release/**"]
 env:
-  GO_VERSION: '1.17'
+  GO_VERSION: '1.18'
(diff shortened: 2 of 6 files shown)


nathancoleman:
Holding until after we release today

mikemorris:
why e2e tests are currently failing:
> TLS 1.0 and 1.1 disabled by default client-side
https://tip.golang.org/doc/go1.18#tls10

nathancoleman (approved):

mikemorris:
For future reference, apparently this missed bumping the actual version used for binary builds, see https://github.com/hashicorp/consul-api-gateway/pull/189

PR #169, opened 2022-04-27
Add promote-dev-docker step to ci.hcl

andrewstucki:
To fix publishing artifacts.


nathancoleman (approved):
Matches https://github.com/hashicorp/crt-core-helloworld/blob/main/.release/ci.hcl#L150

PR #173, opened 2022-04-27
conformance: remove Helm chart consul image config

mikemorris:

- Remove explicit `consul: hashicorp/consul:1.12.0` config from conformance testing Helm config
  - Default was bumped in https://github.com/hashicorp/consul-k8s/pull/1182 and has been included in latest 0.43.0 Helm chart release.

### How I've tested this PR:

Let the conformance tests run in CI on this PR.

Should we switch to pinning against a Helm chart release, or do we like staying on hashicorp/consul-k8s `HEAD`?

### Checklist:
- [ ] ~~Tests added~~
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/testing/conformance/consul-config.yaml
+++ b/internal/testing/conformance/consul-config.yaml
@@ -1,5 +1,4 @@
 global:
-  image: "hashicorp/consul:1.12.0"  # TODO Remove once 1.12.0 is consumed in consul-k8s
   tls:
     enabled: true
 server:


nathancoleman:
>Should we switch to pinning against a Helm chart release, or do we like staying on hashicorp/consul-k8s HEAD?

I was thinking about this last night, and I kinda prefer having our tests run on the bleeding edge so that we find issues early. If you're not super opinionated about it, I'd rather keep it on consul-k8s `main` for now.

PR #177, opened 2022-04-28
Backport assistant config

sarahalsmiller:
Backport assistant workflow

### How I've tested this PR:
Labels on this PR

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,21 @@
+---
+name: Backport Assistant Runner
+
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    container: hashicorpdev/backport-assistant:0.2.5
+    steps:
+      - name: Run Backport Assistant
+        run: backport-assistant backport
+        env:
+          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.x)"
+          BACKPORT_TARGET_TEMPLATE: "release/{{.target}}"
+          GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}


.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    container: hashicorpdev/backport-assistant:0.2.2
+    steps:
+      - name: Run Backport Assistant
+        run: |
+          backport-assistant backport
```
nathancoleman, line 18:
```suggestion
        run: backport-assistant backport
```

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    container: hashicorpdev/backport-assistant:0.2.2
```
nathancoleman, line 14: [Looks like](https://hub.docker.com/r/hashicorpdev/backport-assistant/tags) the latest version is `0.2.5`

```suggestion
    container: hashicorpdev/backport-assistant:0.2.5
```

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - name: Run Backport Assistant
+        run: |
+          backport-assistant backport
+        env:
+          BACKPORT_LABEL_REGEXP: "(?P<target>\\d+\\.\\d+\\.x)-backport"
```
nathancoleman, line 20: It [looks like](https://github.com/hashicorp/consul-api-gateway/labels?q=backport) we've used the inverse of this already - `backport-0.2.x` - but obviously we can create a new standard if we want. Just calling this out

```suggestion
          BACKPORT_LABEL_REGEXP: "backport-(?P<target>\\d+\\.\\d+\\.x)"
```

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+          BACKPORT_TARGET_TEMPLATE: "release/{{.target}}"
```
nathancoleman, line 21: Consul's version of this is slightly different. Should we be doing the same?

https://github.com/hashicorp/consul/blob/892032acae4b2f93e2523afe89d3725310d3755b/.github/workflows/backport-assistant.yml#L52

nathancoleman (changes requested):

.github/workflows/backport.yml
nathancoleman: Ah I'm seeing in the [docs](https://github.com/hashicorp/backport-assistant) that `target` just needs to match what's parsed out of the label. Since we're also parsing the `.x` in our capture group above, we don't need to add the `.x` as Consul does here.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged
```
nathancoleman, line 12: Does this negate the `labeled` trigger above? It'd be sweet if we could label a PR that's already been merged and have the backport run

.github/workflows/backport.yml
mikemorris: Using a shared prefix like `backport` groups the labels together in the selector which is nice. FWIW Consul core uses `backport/` https://github.com/hashicorp/consul/labels

.github/workflows/backport.yml
mikemorris: On Consul core this is possible by adding the label and manually re-running the CI job, which is super-unintuitive to find, would love if this could happen from just adding the label post-merge.

sarahalsmiller: I /think/ this is only checking if the pr has actually been merged before it runs the backport runner, since the merged value is a bool yes/no, which I would argue is probably desirable in that we don't want this to run on PRs that haven't merged yet. If I'm reading everything right, it should run on closed and on labeled, but only run on labeled if the pr has already been merged.

.github/workflows/backport.yml
sarahalsmiller: I basically just copied over the example in the docs, open to doing it the way consul does

.github/workflows/backport.yml
nathancoleman: I can't find a good docs source describing `.merged`; however, from the examples I see in various places, I agree with you @sarahalsmiller

nathancoleman (approved):

PR #186, opened 2022-04-28
Automate version + changelog updates for release

nathancoleman:
Instead of requiring the host of manual updates that we do today, we can just kick off an Actions workflow to make the changes and create a PR for us.

### How I've tested this PR:
Temporarily triggered on push of this branch.
[Example branch this action creates](https://github.com/hashicorp/consul-api-gateway/compare/prepare-release-workflow...v0.3.0-release-prep?expand=1)

> **Note**: I previously had this script creating a PR for the resulting branch; however, Actions is now configured in a way that [blocks PR creation](https://github.blog/changelog/2022-05-03-github-actions-prevent-github-actions-from-creating-and-approving-pull-requests/) 😢 . As a result, you need to run the action and then get the link from the "Create branch" step output.

Review resulting branch linked above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -36,13 +36,6 @@ jobs:
           echo "$VERSION" >> $GITHUB_ENV
           echo "EOF" >> $GITHUB_ENV
 
-      - name: Update version.go
-        uses: jacobtomlinson/gha-find-replace@v2
-        with:
-          include: 'internal/version/version.go'
-          find: 'Version = "[0-9\.]+"'
-          replace: 'Version = "${{ env.NEW_API_GATEWAY_VERSION }}"'
-
       - name: Update deployment.yaml
         uses: 'jacobtomlinson/gha-find-replace@v2'
         with:
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -1,9 +1,6 @@
 name: prepare_release
 
 on:
-  push:
-    branches: ['prepare-release-workflow']
-
   workflow_dispatch:
     inputs:
       new-min-consul-version:
@@ -19,8 +16,8 @@ jobs:
   update-versions:
     runs-on: ubuntu-latest
     env:
-      NEW_CONSUL_REQ: ${{ github.event.inputs.new-min-consul-version || '1.13.0' }}
-      NEW_CONSUL_K8S_REQ: ${{ github.event.inputs.new-min-consul-k8s-version || '0.44.0' }}
+      NEW_CONSUL_REQ: ${{ github.event.inputs.new-min-consul-version }}
+      NEW_CONSUL_K8S_REQ: ${{ github.event.inputs.new-min-consul-k8s-version }}
 
     steps:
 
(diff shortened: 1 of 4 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,130 @@
+name: prepare_release
+
+on:
+  push:
+    branches: ['prepare-release-workflow']
```
nathancoleman, line 5: This is only for demonstration purposes with this PR. Once it merges, we will run this workflow manually from the Actions tab of our repo.

.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,130 @@
...
+        description: 'The new minimum version of Consul K8s (no "v" prefix)'
+        required: true
+        type: string
+
+jobs:
+  update-versions:
+    runs-on: ubuntu-latest
+    env:
+      NEW_CONSUL_REQ: ${{ github.event.inputs.new-min-consul-version || '1.13.0' }}
+      NEW_CONSUL_K8S_REQ: ${{ github.event.inputs.new-min-consul-k8s-version || '0.44.0' }}
```
nathancoleman, line 23: The `|| <value>` is only for the PR workflow mentioned above where no inputs are received. Will remove before merge.

.release/ci.hcl
```diff
@@ -13,8 +13,7 @@ project "consul-api-gateway" {
 }
 
 event "merge" {
-  // "entrypoint" to use if build is not run automatically
-  // i.e. send "merge" complete signal to orchestrator to trigger build
+  // "entrypoint" to use if build is not run automatically i.e. send "merge" complete signal to orchestrator to trigger build
```
nathancoleman, line 16: The hclq tool runs a formatter on this file, so you'll see some changes the first time it runs. After the first time, the diff will just be whatever minor release branch is added below.

Makefile
```diff
@@ -62,7 +62,7 @@ endif
 ifeq (, $(LAST_RELEASE_GIT_TAG))
 	@echo "Please set the LAST_RELEASE_GIT_TAG environment variable to generate a changelog section of notes since the last release."
 else
-	changelog-build -last-release ${LAST_RELEASE_GIT_TAG} -entries-dir .changelog/ -changelog-template .changelog/changelog.tmpl -note-template .changelog/release-note.tmpl -this-release $(shell git rev-parse HEAD)
+	@changelog-build -last-release ${LAST_RELEASE_GIT_TAG} -entries-dir .changelog/ -changelog-template .changelog/changelog.tmpl -note-template .changelog/release-note.tmpl -this-release $(shell git rev-parse HEAD)
```
nathancoleman, line 65: Don't want the command in the output here because we can't pipe it into a CHANGELOG update

nathancoleman (commented):
Personal Review

sarahalsmiller (approved):
Looks solid, I'd love to test run it

mikemorris (approved):
Looks promising!

jeanneryan (approved):
This looks good to me.

PR #190, opened 2022-05-02
ci: update binary build Go version to 1.18.1

mikemorris:
Refs https://github.com/hashicorp/consul-api-gateway/pull/167#issuecomment-1115079861 (which already added a changelog entry for this)


nathancoleman (approved):
Noting that we pinned everything else to the minor version `1.18` in #167 . Why the patch version here?

mikemorris:
Re patch version, we discussed as a team how pinning an exact version makes it easier to see exactly what version a binary was published with when looking through commit history without needing to introspect the actual binary.

PR #195, opened 2022-05-12
Configure instances on gatewayConfig

sarahalsmiller:
- Add optional deploy object to gatewayClassConfiguration that allows users

### How I've tested this PR:
-Created a kind cluster and installed CRD and helmchart locally
- need to test actually creating a gateway and making sure the instances scale appropriately 

- Added builder unit test that makes sure the deployment gets built with the correct number of replicas set
- Maybe need an end to end test to capture the bug I'm seeing where it doesn't actually set the replicas appropriately 

### Checklist:
- [ X ] Tests added
- [ X ] CHANGELOG entry added 


andrewstucki (dismissed):
Biggest change is the need to not reset the replica count if changed out of band. Ideally what we'd do is pass in the count for an existing deployment (if it exists) and use it. If no deployment exists, then do the `defaultInstances`/in-code defaults.

Also, is this PR going to also incorporate the arguments for min/max?

internal/k8s/builder/gateway.go
```diff
@@ -149,18 +150,34 @@ func (b *GatewayDeploymentBuilder) Build() *v1.Deployment {
 	}
 }
 
+func (b *GatewayDeploymentBuilder) instances() *int32 {
```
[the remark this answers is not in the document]
sarahalsmiller: It looks like that might already be getting preserved since it looks like replicas is a layer above template. Might need you to double check on that though.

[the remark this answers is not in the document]
andrewstucki: Ah, interesting, yeah, that would make sense, that's definitely not intended. We should probably audit what isn't handled by this merge operation apart from metadata (which was IIRC the intent of adding this create/update code that bypasses things like metadata).

Think best way to double check this and also make sure we're always allowing for manual modification is to probably add an e2e test for this.

pkg/apis/v1alpha1/types.go
```diff
@@ -51,6 +51,15 @@ type GatewayClassConfigSpec struct {
...
 	// Logging levels
 	LogLevel string `json:"logLevel,omitempty"`
+	// Configuration information about how many instances to deploy
+	DeploySpec DeploySpec `json:"deploy,omitempty"`
+}
+
+// +k8s:deepcopy-gen=true
+
+type DeploySpec struct {
+	//Number of gateway instances that should be deployed
```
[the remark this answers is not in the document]
sarahalsmiller: This jsonapi? https://github.com/google/jsonapi I'm not opposed to trying it, just don't see it anywhere else in the code base at the moment.

internal/k8s/builder/gateway.go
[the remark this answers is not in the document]
sarahalsmiller: Added two e2e tests, one for normal replication and one for testing to make sure we can modify the replica number manually. Let me know your thoughts.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -123,14 +124,102 @@ func TestGatewayWithClassConfigChange(t *testing.T) {
...
+			}, 60*time.Second, checkInterval, "no gateway found in the allotted time")
+
+			require.Eventually(t, gatewayStatusCheck(ctx, resources, gatewayName, namespace, conditionReady), 30*time.Second, checkInterval, "no gateway found in the allotted time")
+			checkGatewayConfigAnnotation(t, gateway, gatewayClassConfig)
+
+			deployment := &appsv1.Deployment{}
+			assert.NoError(t, resources.Get(ctx, gatewayName, namespace, deployment))
+			assert.Equal(t, initialReplicas, *deployment.Spec.Replicas)
+
+			deployment.Spec.Replicas = &finalReplicas
```
nathancoleman, line 202: I find the one-liner comments above to be super helpful in grepping what a test is doing. Would you mind adding one here?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -123,14 +124,102 @@ func TestGatewayWithClassConfigChange(t *testing.T) {
...
+			require.Eventually(t, func() bool {
+				err := resources.Get(ctx, gatewayName, namespace, gateway)
+				return err == nil && conditionReady(gateway.Status.Conditions)
+			}, 60*time.Second, checkInterval, "no gateway found in the allotted time")
+
+			require.Eventually(t, gatewayStatusCheck(ctx, resources, gatewayName, namespace, conditionReady), 30*time.Second, checkInterval, "no gateway found in the allotted time")
+			checkGatewayConfigAnnotation(t, gateway, gatewayClassConfig)
+
+			deployment := &appsv1.Deployment{}
+			assert.NoError(t, resources.Get(ctx, gatewayName, namespace, deployment))
```
nathancoleman, line 199: Loving your distinction between `require` and `assert` ❤️

pkg/apis/v1alpha1/types.go
```diff
@@ -193,9 +205,15 @@ func compareServices(a, b *corev1.Service) bool {
 // MergeDeploymentmerges a gateway deployment a onto b and returns b, overriding all of
```
nathancoleman, line 205:
```suggestion
// MergeDeployment merges a gateway deployment a onto b and returns b, overriding all of
```

pkg/apis/v1alpha1/types.go
```diff
@@ -193,9 +205,15 @@ func compareServices(a, b *corev1.Service) bool {
...
 // the fields that we'd normally set for a service deployment. It does not attempt
 // to change the service type
-func MergeDeployment(a, b *appsv1.Deployment) *appsv1.Deployment {
+func MergeDeployment(a, b *appsv1.Deployment, currentReplicas *int32) *appsv1.Deployment {
 	if !compareDeployments(a, b) {
 		b.Spec.Template = a.Spec.Template
+
+		//Dont override if replicas have already been set on deployment
+		//TODO this logic will likely need to be updated as part of CAPIGW-137
```
nathancoleman, line 213:
```suggestion
		// Don't override if replicas have already been set on deployment
		// TODO this logic will likely need to be updated as part of CAPIGW-137
```

internal/k8s/reconciler/gateway.go
```diff
@@ -464,8 +464,19 @@ func (g *K8sGateway) ensureDeploymentExists(ctx context.Context) error {
...
+	currentDeployment, err := g.client.GetDeployment(ctx, types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name})
+	if err != nil {
+		return err
+	}
+
+	var currentReplicas *int32
+	if currentDeployment != nil {
+		currentReplicas = currentDeployment.Spec.Replicas
+	}
+	mutated = apigwv1alpha1.MergeDeployment(deployment, mutated, currentReplicas)
```
nathancoleman, line 477: This is existing code, but it doesn't make sense to me why we "merge" a freshly-generated deployment onto a `.DeepCopy()` of itself.

This drew my attention because we're passing information from a third `Deployment` we have in memory as `currentReplicas`, which strikes me as odd.

Based on the context, I would expect `MergeDeployment(deployment, currentDeployment)` so that we're merging a newly-constructed deployment onto the existing one.

Can you speak to this @andrewstucki ?

pkg/apis/v1alpha1/types.go
```diff
@@ -223,5 +241,10 @@ func compareDeployments(a, b *appsv1.Deployment) bool {
 			}
 		}
 	}
+
+	if b.Spec.Replicas != a.Spec.Replicas {
```
nathancoleman, line 245: Curious if this is comparing values or addresses

nathancoleman (changes requested):
Looking good! One big question about why pre-existing code is merging a deployment onto a copy of itself 🤔 . That's probs for Andrew though.

config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
```diff
@@ -90,6 +90,19 @@ spec:
                       type: string
                     type: array
                 type: object
+              deploy:
```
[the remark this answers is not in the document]
sarahalsmiller: Sounds good to me!

pkg/apis/v1alpha1/types.go
sarahalsmiller: Double checked it and it compares addresses, good catch. Fixed.

internal/k8s/builder/testdata/multiple-instances.deployment.golden.yaml
```diff
@@ -0,0 +1,96 @@
+metadata:
+  creationTimestamp: null
+  labels:
+    api-gateway.consul.hashicorp.com/created: "-62135596800"
+    api-gateway.consul.hashicorp.com/managed: "true"
+    api-gateway.consul.hashicorp.com/name: test-multiple-instances
+    api-gateway.consul.hashicorp.com/namespace: ""
+  name: test-multiple-instances
+spec:
+  replicas: 10
```
nathancoleman, line 10: Should we expect an `8` here now?

internal/k8s/builder/testdata/multiple-instances.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: api-gateway.consul.hashicorp.com/v1alpha1
+kind: GatewayClassConfig
+metadata:
+  name: test-gateway-class-config
+spec:
+  image:
+    consulAPIGateway: hashicorp/consul-api-gateway:0.2.1
+  serviceType: "ClusterIP"
+  deployment:
+    defaultInstances: 10
```
nathancoleman, line 10: Should we expect an `8` here now?

sarahalsmiller: Good point. I took out the logic that's actually checking it since I added the max value to the kube validation, but I'll update the test for consistencies sake.

internal/k8s/reconciler/gateway.go
sarahalsmiller: For what its worth, in https://github.com/hashicorp/consul-api-gateway/pull/198 I pulled the logic out of mergedeployment and put it into the build function. I still don't know why this is needed though.

andrewstucki: So, this could probably be refactored in general, but the idea is that the merge happens on an update, identically to the controller runtime helper that does the same sort of thing: https://github.com/hashicorp/consul-api-gateway/blob/9d0b1d8de9db535edf0c6e8ad33feddba4feb0b4/internal/k8s/gatewayclient/gatewayclient.go#L368

if you dig into the actual implementation of that helper, basically when it detects a change in what exists in k8s v. what exists on disk, it overrides the passed in object (`mutated` in this case), then calls the mutation functions before calling `client.Update`. The `DeepCopy` is done so that on:

1. `Create`, the object gets created as it should, and 
2. On `Update`, when `MergeDeployment` is called, the value is mutated to be reflective of the constructed state.

What this means is that @sarahalsmiller you'll need to put the `MergeDeployment` call back inside of the callback -- that's where any mutations should happen for an `Update` call.

pkg/apis/v1alpha1/types.go
```diff
@@ -51,6 +51,18 @@ type GatewayClassConfigSpec struct {
...
+}
+
+// +k8s:deepcopy-gen=true
+
+type DeploymentSpec struct {
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Maximum=8
+	// +kubebuilder:validation:Minimum=1
+	//Number of gateway instances that should be deployed by default
+	Instances *int32 `json:"defaultInstances,omitempty"`
```
nathancoleman, line 65: I would prefer to have the name here match the yaml config to remove some mental overhead, but that's just preference

internal/k8s/reconciler/gateway.go
```diff
@@ -464,8 +464,19 @@ func (g *K8sGateway) ensureDeploymentExists(ctx context.Context) error {
...
+	currentDeployment, err := g.client.GetDeployment(ctx, types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name})
+	if err != nil {
+		return err
+	}
+
+	var currentReplicas *int32
+	if currentDeployment != nil {
+		currentReplicas = currentDeployment.Spec.Replicas
+	}
+
```
nathancoleman, line 477: I think a comment explaining why we're fetching the deployment and replica count would be useful

pkg/apis/v1alpha1/types.go
```diff
@@ -51,6 +51,18 @@ type GatewayClassConfigSpec struct {
...
+	DeploymentSpec DeploymentSpec `json:"deployment,omitempty"`
+}
+
+// +k8s:deepcopy-gen=true
+
+type DeploymentSpec struct {
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Maximum=8
+	// +kubebuilder:validation:Minimum=1
+	//Number of gateway instances that should be deployed by default
```
nathancoleman, line 64:
```suggestion
	// Number of gateway instances that should be deployed by default
```

nathancoleman (approved):

PR #197, opened 2022-05-17
store: remove route from listener when not able to bind

mikemorris:
Removes a route from a listener if it fails to bind when reconciling. Fixes an issue where deleting a ReferencePolicy would update the route status, but the stale route would not be removed from the gateway listener.

As an aside, the `(bool, error)` return of `CanBind` seems redundant and makes it prone to misuse - should it be changed to just one of those values instead of the tuple?

### How I've tested this PR:
- [x] Add checks for gateway listener status `attachedRoutes` after deleting ReferencePolicy in e2e test
- [x] Add checks for TCP and HTTP connection errors after deleting ReferencePolicy in e2e test
- [x] Add e2e test for switching route binding from one gateway to another

- Review the test additions/changes.
- Confirm that a gateway removes a stale route and stops routing traffic after a route is invalidated by being modified or having a ReferencePolicy that affects it deleted.
- The `checkRouteError` function and only looking for an empty string TCP response feel like quite crude checks, is there a better way to handle these?
- `TestRouteParentRefChange` contains the checks/assertions that are _currently_ working as expected - I believe there's a _different_ bug here still preventing some functionality from working as expected, but punted further exploration of that out to https://github.com/hashicorp/consul-api-gateway/pull/200
- The debug logging I added in the helper funcs was quite helpful and I'm leaning towards wanting to keep it somehow, should it be switched to something like `t.Log` instead though?

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/hashicorp/consul-api-gateway
 
-go 1.17
+go 1.18
 
 require (
 	github.com/armon/go-metrics v0.3.9
@@ -18,6 +18,7 @@ require (
 	github.com/prometheus/client_golang v1.11.0
 	github.com/stretchr/testify v1.7.0
 	github.com/vladimirvivien/gexe v0.1.1
+	golang.org/x/exp v0.0.0-20220518171630-0b5c67f07fdf
 	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	google.golang.org/grpc v1.40.0
@@ -99,7 +100,7 @@ require (
 	go.uber.org/zap v1.19.0 // indirect
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2 // indirect
 	golang.org/x/oauth2 v0.0.0-20210402161424-2e8d93401602 // indirect
-	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
+	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654 // indirect
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d // indirect
 	golang.org/x/text v0.3.6 // indirect
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac // indirect
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"net/http"
 	"os"
@@ -1789,20 +1788,20 @@ func checkRoute(t *testing.T, port int, path, expected string, headers map[strin
 		}
 
 		resp, err := client.Do(req)
-		log.Print("RESPONSE_ERR: ", err)
 		if err != nil {
+			t.Log(err)
 			return false
 		}
 		defer resp.Body.Close()
 
 		data, err := io.ReadAll(resp.Body)
-		log.Print("READALL_DATA: ", string(data))
-		log.Print("READALL_ERR: ", err)
 		if err != nil {
+			t.Log(err)
 			return false
 		}
+		t.Log(string(data))
 
-		log.Print("STATUS_CODE: ", resp.StatusCode)
+		t.Log("status code", resp.StatusCode)
 		if resp.StatusCode != http.StatusOK {
 			return false
 		}
@@ -1845,15 +1844,16 @@ func checkTCPRoute(t *testing.T, port int, expected string, exact bool, message 
 			Port: port,
 		})
 		if err != nil {
-			log.Print("TCP_DIAL_ERROR: ", err)
+			t.Log(err)
 			return false
 		}
 		data, err := io.ReadAll(conn)
 		if err != nil {
-			log.Print("TCP_READALL_ERROR: ", err)
+			t.Log(err)
 			return false
 		}
-		log.Print("TCP_READALL_DATA: ", string(data))
+		t.Log(string(data))
+
 		if exact {
 			return string(data) == expected
 		} else {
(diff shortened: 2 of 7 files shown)


nathancoleman:
@mikemorris is this ready for review? wasn't sure if you expect tests to be failing

mikemorris:
@nathancoleman Not quite yet, finishing up the last e2e test now - the test failures (aside from changelog) look unrelated and unexpected though?

nathancoleman:
Those are interesting... Gonna rerun and see what happens 🤞

mikemorris:
Would appreciate reviews on this now - there are a few outstanding questions in FIXME and TODO comments, and I think some Consul sync adapter logic that needs to be updated to properly handle when the last route is removed from a gateway.

mikemorris:
This is cleaned up and updated the PR reviewer guidance section and ready for a review again.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/consul-api-gateway
 
-go 1.17
+go 1.18
```
nathancoleman, line 3: Good catch!

internal/store/memory/gateway.go
```diff
@@ -60,6 +60,7 @@ func (g *gatewayState) TryBind(ctx context.Context, route store.Route) {
 			if err != nil {
 				// consider each route distinct for the purposes of binding
 				g.logger.Debug("error binding route to gateway", "error", err, "route", route.ID())
+				l.RemoveRoute(route.ID())
```
nathancoleman, line 63: Will we always get an `err` value when `canBind == false`? If not,  should we also  call `RemoveRoute` below when `!canBind`?

nathancoleman: > As an aside, the (bool, error) return of CanBind seems redundant and makes it prone to misuse - should it be changed to just one of those values instead of the tuple?

I would **much** prefer a single return value if `canBind` is always `true` unless `err != nil`.

internal/testing/e2e/consul.go
```diff
@@ -83,21 +83,23 @@ func init() {
...
+	httpFlattenedPort                int
+	httpReferencePolicyPort          int
+	tcpReferencePolicyPort           int
+	parentRefChangeFirstGatewayPort  int
+	parentRefChangeSecondGatewayPort int
+	grpcPort                         int
+	extraHTTPPort                    int
+	extraTCPPort                     int
+	extraTCPTLSPort                  int
+	extraTCPTLSPortTwo               int
```
nathancoleman, line 100: ~~Would it be possible to just accept a list of `portsToOpen []int`?~~

Looks like we need a specific reference to each port type for testing, not just to open ports on Kind

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1694,13 +1974,20 @@ func checkTCPRoute(t *testing.T, port int, expected string, message string) {
 			Port: port,
 		})
 		if err != nil {
+			log.Print("TCP_DIAL_ERROR: ", err)
```
nathancoleman, line 1977: Were these `Prints` intended to stick around?

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1394,6 +1435,191 @@ func TestReferencePolicyLifecycle(t *testing.T) {
...
+			// Check that route unbinds from first gateway listener successfully
+			require.Eventually(t, func() bool {
+				updated := &gateway.HTTPRoute{}
+				if err := resources.Get(ctx, httpRouteName, namespace, updated); err != nil {
+					return false
+				}
+				firstGatewayParentRefFound := false
+				for _, status := range updated.Status.Parents {
+					if string(status.ParentRef.Name) == firstGatewayName {
+						firstGatewayParentRefFound = true
```
nathancoleman, line 1608: Could just `return false` here, no need to check the rest

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1394,6 +1435,191 @@ func TestReferencePolicyLifecycle(t *testing.T) {
 	testenv.Test(t, feature.Feature())
 }
 
+func TestRouteParentRefChange(t *testing.T) {
```
nathancoleman, line 1438: I like this test! Nice work

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1171,15 +1168,35 @@ func TestReferencePolicyLifecycle(t *testing.T) {
...
+							From: &fromSelector,
+							Selector: &meta.LabelSelector{
+								MatchExpressions: []meta.LabelSelectorRequirement{{
+									Key:      "kubernetes.io/metadata.name",
+									Operator: "In",
+									Values:   []string{httpRouteNamespace},
+								}},
+							},
+						},
+					},
```
nathancoleman, line 1182: 🏆

nathancoleman (commented):
Not a fan of `CanBind` returning two values, as you mentioned in the PR body. Would love to simplify that. Otherwise, just a couple of non-functional questions and LGTM!

internal/commands/server/k8s_e2e_test.go
nathancoleman, line 1608:
```suggestion
						return false
```

internal/commands/server/k8s_e2e_test.go
mikemorris: I found them quite useful when writing the tests (and wasted time fiddling around with like three different print functions to find something that would actually show up in test output), had an open question in PR description about keeping these or changing to `t.Log`

internal/testing/e2e/consul.go
mikemorris: Yea, all this boilerplate is a significant pain point in writing new e2e tests 😞

internal/commands/server/k8s_e2e_test.go
nathancoleman: Whoops, I missed that one - my bad. I'm fine leaving them if they're helpful but would switch to `t.Log[f]`

internal/store/memory/gateway.go
mikemorris: Started trying to rip this out and ran into the following bit which does seem to return `false, nil` in some cases https://github.com/hashicorp/consul-api-gateway/blob/2362a739269ac7778ed41a8772602d4914dfe553/internal/k8s/reconciler/listener.go#L339-L388

nathancoleman (approved):

PR #199, opened 2022-05-26
e2e: fixup checkGatewayConfigAnnotation and clean up TestGatewayWithClassConfigChange

mikemorris:

- Refactor `checkGatewayConfigAnnotation` to get the gateway resource directly rather than relying on passing in an updated gateway resource as a side effect of a prior check.
- Fixes an incorrect assertion and removes substantial redundant checks from TestGatewayWithClassConfigChange
- Minor refactor to pass `resources.Resources` instead of `*envconf.Config` into `createGatewayClass` and `createGateway` helpers, clean up many tests to use these helpers when possible.

The motivation for this was stumbling over writing a new test with a similar two gateway setup - finding several redundant checks and spending too long figuring out why things weren't working as expected when deleting the verbose checks in favor of the streamlined helpers (the subsequent `checkGatewayConfigAnnotation` checks were dependent on a side effect from the verbose status check).

### How I've tested this PR:

Running the e2e tests. The following snippet was helpful for iterating more quickly:
```
go test ./... -tags e2e -v -failfast -run TestGatewayWithClassConfigChange
```

Confirm that these changes seem desirable.

### Checklist:
- [x] Tests added

### DIFF
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -927,7 +927,6 @@ func TestReferencePolicyLifecycle(t *testing.T) {
 
 			resources := cfg.Client().Resources(namespace)
 
-			// Create a GatewayClassConfig
 			_, gc := createGatewayClass(ctx, t, cfg, 1)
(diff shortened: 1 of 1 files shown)


nathancoleman (approved):
Beautiful!

PR #200, opened 2022-05-26
fix listener and route behavior when changing ParentRef

mikemorris:

Fixes the following issue:
- [x] Prior gateway listener doesn't update `attachedRoutes` back to zero when last route removes a ParentRef for that gateway

Updates test to document currently-expected behavior but not fixed yet - after discussing with @nathancoleman we believe support for multiple gateways is not yet implemented rather than a "bug":
- [x] An HTTPRoute fails to route traffic from a second ParentRef gateway to a backend service (503 reponses with `no healthy upstream`) until the first ParentRef is removed
  - Traffic _does_ still route successfully through the first gateway after adding the second gateway ParentRef

### How I've tested this PR:

Written e2e test assertions that demonstrate broken behavior when changing the ParentRef of a route.

Confirm that new test assertions demonstrate intended functionality, review fixes.

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/200.txt
+++ b/.changelog/200.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-Clean up stale routes from gateway listeners which no longer reference the gateway.
+Clean up stale routes from gateway listeners when route no longer references the gateway.
 ```
--- a/.changelog/200.txt
+++ b/.changelog/200.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Clean up stale routes from gateway listeners which no longer reference the gateway.
+```
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -1374,6 +1374,9 @@ func TestRouteParentRefChange(t *testing.T) {
 				listenerAttachedRoutes(1),
 			), checkTimeout, checkInterval, "listeners not ready in the allotted time")
 
+			// Check that HTTPRoute is successfully resolved and routing traffic
+			checkRoute(t, secondGatewayCheckPort, "/", serviceOne.Name, nil, "service one not routable in allotted time")
+
 			// Check that route unbinds from first gateway listener successfully
 			require.Eventually(t, func() bool {
 				updated := &gateway.HTTPRoute{}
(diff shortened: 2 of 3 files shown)


internal/commands/server/k8s_e2e_test.go
```diff
@@ -1388,6 +1391,19 @@ func TestRouteParentRefChange(t *testing.T) {
 				return true
 			}, checkTimeout, checkInterval, "HTTPRoute status not unset in allotted time")
 
+			// FIXME: why are attached routes not being pruned as expected in this case?
```
mikemorris, line 1394:
```suggestion
```

.changelog/200.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Clean up stale routes from gateway listeners which no longer reference the gateway.
```
mikemorris, line 2:
```suggestion
Clean up stale routes from gateway listeners when route no longer references the gateway.
```

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1374,6 +1416,28 @@ func TestRouteParentRefChange(t *testing.T) {
...
+
+			// TODO: Routing from multiple gateways is not yet supported.
+			// When implemented, this check should be updated to wait for
+			// http.StatusOK and serviceOne.Name as the body content prefix.
+			checkRoute(t, secondGatewayCheckPort, "/", httpResponse{
+				StatusCode: http.StatusServiceUnavailable,
+				Body:       "no healthy upstream",
+			}, nil, "service one not returning expected error from second gateway in allotted time")
+
+			// Update httpRoute from remote, then remove first gateway ParentRef
```
nathancoleman, line 1433: What do you mean when saying `... from remote`?

nathancoleman (approved):

internal/commands/server/k8s_e2e_test.go
mikemorris: Without a `resources.Get` call to update the local variable here, the subsequent call to `resources.Update` will fail because the version is lower than the version of that object in the cluster (due to gateway controller settings status conditions I believe).

nathancoleman: Oh so `from remote` here is just pull down an updated copy from the kube API

PR #201, opened 2022-05-27
Multiple Pod IPs

sarahalsmiller:
- Assign multiple ips to gateways when there are multiple pods

### How I've tested this PR:
- Unit tests passing
- attempting to replicate circumstances on a kind cluster

- Unit tests passing 

### Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/reconciler/gateway.go
```diff
@@ -221,22 +221,26 @@ func (g *K8sGateway) assignGatewayIPFromService(ctx context.Context, service *co
...
 		return nil
 	}
 
-	if pod.Status.PodIP != "" {
-		g.serviceReady = true
-		g.addresses = append(g.addresses, pod.Status.PodIP)
+	for _, pod := range pods {
+		if pod.Status.PodIP != "" {
+			g.serviceReady = true
+			if !ipInList(g.addresses, pod.Status.PodIP) {
```
nathancoleman, line 240: Is the check for duplicate IPs in list necessary? I think `g.addresses` is built up from scratch here every time, and the k8s API shouldn't return the same pod twice

sarahalsmiller: I can remove it, I mostly just added it because I already was adding it for the HostIP section

mikemorris (approved):
LGTM (just some additional minor `pod` -> `pods` renaming) - it appears the existing `internal/k8s/reconciler/gateway_test.go` tests are fully exercising that the expected host or pod IPs are being set on the gateway.

Probably worth adding a changelog entry?

.changelog/201.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:note
+Gateway ip address assignment logic updated to include the case when multiple different pod ips exist
```
nathancoleman, line 2:
```suggestion
Gateway IP address assignment logic updated to include the case when multiple different pod IPs exist
```

nathancoleman (approved):

PR #202, opened 2022-05-31
Anti Affinity rules added to deployment spec

sarahalsmiller:
- Added anti-affinity rule to deployment spec to prevent multiple 

### How I've tested this PR:
Built a multinode GCP cluster and making sure build schedules nodes appropriately 
- Pods avoid being scheduled on the same node, but if replicas > available nodes the schedule schedules regardless
Unit tests passing

- Unit tests passing

### Checklist:
- [ X] Tests added
- [ ] CHANGELOG entry added 


nathancoleman:
@sarahalsmiller should probs include a release note here

internal/k8s/builder/gateway.go
```diff
@@ -190,7 +190,42 @@ func (b *GatewayDeploymentBuilder) podSpec() corev1.PodSpec {
...
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					{
+						Weight: 3,
+						PodAffinityTerm: corev1.PodAffinityTerm{
+							LabelSelector: &metav1.LabelSelector{
+								MatchLabels: labels,
+							},
+							TopologyKey: `topology.kubernetes.io/region`,
+						},
+					},
```
nathancoleman, line 207: @mikemorris raised a good point in Slack:

> spread across regions may be an unlikely/undesirable deployment due to cross-region data transfer costs and latency - it may be okay for it to be ranked first if deployed like that, but likely not something we would recommend (and iirc we won't even technically support it until agentless work lands due to gossip latency requirements)

Probably best to drop this guy and just go zone -> host.

internal/k8s/builder/gateway.go
```diff
@@ -190,7 +190,42 @@ func (b *GatewayDeploymentBuilder) podSpec() corev1.PodSpec {
...
+					{
+						Weight: 2,
+						PodAffinityTerm: corev1.PodAffinityTerm{
+							LabelSelector: &metav1.LabelSelector{
+								MatchLabels: labels,
+							},
+							TopologyKey: `topology.kubernetes.io/zone`,
```
nathancoleman, line 214: I was disappointed I couldn't find any constants in the k8s libraries for this, but probably useful to have a constant if only for testing purposes

internal/k8s/builder/gateway.go
nathancoleman: After some discussion with the team, we've decided to ship just the `host` affinity term for now

.changelog/202.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:note
```
nathancoleman, line 1:
```suggestion
```release-note:feature
```

.changelog/202.txt
```diff
@@ -0,0 +1,3 @@
...
+Anti-affinity rules defined so the scheduler will attempt to evenly spread gateway pods across all available nodes
```
nathancoleman, line 2:
```suggestion
Define anti-affinity rules so that the scheduler will attempt to evenly spread gateway pods across all available nodes
```

nathancoleman (changes requested):

nathancoleman (approved):

PR #205, opened 2022-06-01
conformance: skip failing tests

mikemorris:
Implement our own test runner for the conformance test suite that has a known list of tests to skip. This substantially changes how we approach patching the base manifests now that we're not running the conformance suite from within a local checkout - we apply the kustomzie patch using the `krusty` library instead of using the CLI.

> **Note**: our conformance tests are failing on a couple of fronts right now:
> - New assertions were added that require us to return a 404 when a rule doesn’t match our request, either because one isn’t configured or because one failed to attach. We can’t currently program Envoy to return a 404 in these cases and so are returning the default 503.
> - The conformance tests for ReferencePolicy wound up with some different behavior expectations based on discussion with others upstream. We want to skip these tests temporarily until we implement the updated logic.
> - Two tests which create gateways as part of the tests rather than through the base manifests (`HTTPRouteListenerHostnameMatching` and `HTTPRouteDisallowedKind`) are currently hanging indefinitely, and so have been temporarily skipped until we can track down the cause of that issue.

### How I've tested this PR:
- Conformance tests are now passing in CI again (with a few skipped currently that need followup triage).

### Notes for reviewers
- The `go.mod` updates to k8s.io modules are due to adding the dependency on `kyaml` which switches from `googleapis/gnostic` to `google/gnostic` - this was handled in https://github.com/kubernetes/kubernetes/pull/108994 to resolve the dependency conflict, but requires updating all dependencies in lockstep
- The `internal/k8s/logger.go` changes are to adapt to the breaking change in https://github.com/go-logr/logr/pull/42 hit due to the dependency bump to go-logr/logr from updating the Kubernetes ecosystem libraries
- This currently depends on our fork to allow reading manifests from the local filesystem instead of just the read-only embedded filessystem https://github.com/kubernetes-sigs/gateway-api/compare/master...hashicorp:conformance/get-contents-from-path-or-url-local
  - See https://github.com/kubernetes-sigs/gateway-api/issues/1180

- Decide if we prefer to install Consul via the latest commit on the `main` branch of the Helm chart source (catches breaking changes earlier, requires keeping `global.imageK8S: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s-control-plane:latest"` in `consul-config.yaml`) or latest published release from the `hashicorp` Helm repository (incompatible with specifying `latest` tag for `global.imageK8s`)
- Can we PR something upstream to allow setting `maxTimeToConsistency` in the `suite.go` file we're reimplementing instead of [relying on a fork](https://github.com/kubernetes-sigs/gateway-api/compare/master...hashicorp:60s-time-to-consistency)?

### Checklist:
- [x] Tests ~~added~~ skipped


mikemorris (commented):
Late reviewing but LGTM

mikemorris:
Well this is now compiling/linting again after all the restructuring and dependency updates, but seems to be getting stuck spinning up the base manifests resources (due to resource constraints?)

mikemorris:
After chatting with @nathancoleman realized that the kustomization wasn't actually being applied (needed to write the output back to disk and pass that path as `BaseManifests` config, `krusty` doesn't mutate in place). With that done, a few tests skipped for future triage, and a hopefully short-lived workaround for loading manifests from local filesystem pending https://github.com/kubernetes-sigs/gateway-api/issues/1180, conformance tests are now passing again!

nathancoleman:
@mikemorris I'm a little fuzzy on this one. Is it something we still need to merge or should it be closed?

mikemorris:
A substantial chunk of this was implemented in #224, the rest is blocked on https://github.com/kubernetes-sigs/gateway-api/pull/1198 currently, probably makes most sense to close this and open a fresh PR later.

PR #206, opened 2022-06-02
Trigger CRT workflows for main branch

nathancoleman:
- Add `main` to the list of branches that trigger Common Release Tooling workflows. This will produce the `dev_tags` images of consul-api-gateway defined [here](https://github.com/hashicorp/consul-api-gateway/blob/c8dd3efb17fd62091d3baaac956380728f87a3d9/.github/workflows/build.yml#L120) onto hashicorppreview whenever we push commits to `main`. Matches what Consul is doing [here](https://github.com/hashicorp/consul/blob/74ca6406ea7569ad42e5c8ffb575b42e3e524238/.release/ci.hcl#L13).
- Change CRT notification target to different Slack channel where the additional noise created by merges to `main` won't interrupt our daily conversations.
- Publish mutable + immutable `dev_tags` so that we can always test with a specific image even after newer versions have been published. I used the naming patterns from Consul's hashicorppreview repo [here](https://hub.docker.com/r/hashicorppreview/consul/tags).
- Change `version.go` to reflect our next release, `0.3.0`, so that dev tags reflect the in-progress work correctly

### How I've tested this PR:
There's not a good way to test this change without having the modified actions file merged into `main`, but the stakes are very low since we're only modifying our own internal process.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,8 +109,10 @@ jobs:
       - name: Remove + from version
         run: |
           echo "version=$(echo $version | sed 's/+ent//g')" >> $GITHUB_ENV
-      - name: Docker Build
+
+      - name: Docker Build (Release Branch)
         uses: hashicorp/actions-docker-build@v1
+        if: startsWith(github.ref, 'refs/heads/release/')
         with:
           version: ${{env.version}}
           target: default
@@ -118,5 +120,17 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
           dev_tags: |
-            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}-dev
-            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}-dev-${{github.sha}}
+            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}
+
+      - name: Docker Build (Main Branch)
+        uses: hashicorp/actions-docker-build@v1
+        if: github.ref == 'refs/heads/main'
+        with:
+          version: ${{env.version}}
+          target: default
+          arch: ${{matrix.arch}}
+          tags: |
+            docker.io/hashicorp/${{env.repo}}:${{env.version}}
+          dev_tags: |
+            docker.io/hashicorppreview/${{env.repo}}:main
+            docker.io/hashicorppreview/${{env.repo}}:${{github.sha}}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,8 @@ name: build
 
 on:
   push:
-    branches:    
+    branches:
+      - 'main'
       - 'release/**'
 
 env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,9 +110,8 @@ jobs:
         run: |
           echo "version=$(echo $version | sed 's/+ent//g')" >> $GITHUB_ENV
 
-      - name: Docker Build (Release Branch)
+      - name: Docker Build
         uses: hashicorp/actions-docker-build@v1
-        if: startsWith(github.ref, 'refs/heads/release/')
         with:
           version: ${{env.version}}
           target: default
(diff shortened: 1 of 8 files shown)


sarahalsmiller (approved):
Does it matter that the -dev tag will still be applied to preview builds in our release branches? Otherwise looks fine to me.

nathancoleman:
I don't think it particularly matters, but it might get kinda gross. I might leave as-is for release branches and fork to publish `consul-api-gateway:main-latest, consul-api-gateway:main-<sha>` for the `main` branch. Thoughts?

That way each release branch is publishing it's version, but we aren't publishing from main branch as `<most-recent-version>-dev` which can be kinda confusing

mdeggies:
The consul PR was recently updated. Just want clarify the names of tags as the PR currently stands--

PR's merged to `release/0.2.x` will trigger artifact builds, including docker dev builds. Tags will be called:

- `hashicorppreview/consul-api-gateway:0.2.1-dev`. There's [no prerelease string set](https://github.com/hashicorp/consul-api-gateway/blob/release/0.2.x/internal/version/version.go#L22) on this release branch, but since you added the static `-dev` prefix to the tag in `dev_tags`, that will be added a suffix. This tag will be updated every time a new PR is merged. 
- `hashicorppreview/consul-api-gateway:0.2.1-dev-$GITSHA`. This tag will be static and will never be updated since it points to a specific git commit sha. 

On merges to `main`, since the [version string](https://github.com/hashicorp/consul-api-gateway/blob/main/internal/version/version.go#L17) is also set to `0.2.1` with no prerelease, `hashicorppreview/consul-api-gateway:0.2.1-dev` will be updated to the head of `main`.

.github/workflows/build.yml
```diff
@@ -117,4 +118,5 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
           dev_tags: |
-            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}
+            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}-dev
+            docker.io/hashicorppreview/${{env.repo}}:${{env.version}}-dev-${{github.sha}}
```
mikemorris, line 122: Could we slice off the patch segment so this is just `MAJOR.MINOR-dev` to reflect the intended change in https://github.com/hashicorp/consul/pull/13308?

mdeggies:
Whoops, I hit send too quickly. We need to document this better, but dropping the patch version from the release branches is an option. The tags from release branches will then be `<major>.<minor>-dev` and `<major>.<minor>-$gitsha`, and `main` would have the patch version.

mikemorris (approved):
LGTM with caveat that we may want to tweak the dev tag format slightly. +1 to doing _something_ different with the `main` branch so we don't end up with a misleading tag name (having `0.2.1-dev` represent potential `0.3.0` prerelease work while `0.2-dev` accurately represents an upcoming `0.2.2` release feels undesirable), not sure if there's any existing convention around just like a plain `dev` and `dev-$GITSHA` tag instead of something verbose/redundant like `main-latest`?

mikemorris:
@mdeggies Should we consider keeping `VersionPrerelease = "dev"` rather than appending this tag in CI config?

.github/workflows/build.yml
nathancoleman: I love this idea; however, my focus this PR is just getting the `main` branch published to unblock other work.

Mind if we circle back with a targeted PR to change how we're tagging for `release/*` branches?

nathancoleman:
I've differentiated the main branch `dev_tags` so that we don't wind up with misleading tags including version numbers.

- Commits on `release/*` branches will be tagged as `hashicorppreview/consul-api-gateway:<version>`
- Commits on `main` branch will be tagged as both:
  - `hashicorppreview/consul-api-gateway:main` which will point to the latest build from `main`
  - `hashicorppreview/consul-api-gateway:<commit_sha>` which will forever point to the build for the specific commit

@sarahalsmiller @mdeggies @mikemorris can you take a look and see what ya think?

nathancoleman:
Discussed with @mdeggies a bit ago and have a plan for additional changes to fit RelEng's new guidelines. The main part will be a workflow change for our team. More to come, but draft for now

mdeggies (approved):
This looks great Nathan! 🥇

PR #211, opened 2022-06-06
fix codeowners

modrake:
I made a mistake when changing CODEOWNERS to lock down release changes in this repo a few days back, this fixes my mistake.


claire-labry (approved):
:shipit:

PR #216, opened 2022-06-08
ci: remove broken Consul binary install

mikemorris:

- Remove flaky Consul install using https://github.com/nickethier/action-setup-hashicorp-tool
- Install binary directly with `curl` instead

### How I've tested this PR:

- Let CI tests run and confirm they don't fail on downloading Consul binary

- Confirm that CI tests are less flaky with this approach

### Checklist:
- [ ] ~~Tests added~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,12 +50,6 @@ jobs:
       with:
         go-version: ${{ env.GO_VERSION }}
 
-    - name: Install Consul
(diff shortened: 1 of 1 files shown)


nathancoleman (approved):
Hey! Look at those unit test runs passing on the first go! 🏆 

This matches the alternative being used [here](https://github.com/hashicorp/consul-ecs/blob/d282670fc240cb645d90bf3f9563b711e1d616be/.github/workflows/bin-ci.yml#L60-L68).

PR #221, opened 2022-06-09
Update builds to Go 1.18.3

nathancoleman:
Update builds from Go `1.18.1` to `1.18.3`

### How I've tested this PR:
🤖 tests pass

🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
       matrix:
         goos: ["linux"]
         goarch: ["arm", "arm64", "386", "amd64"]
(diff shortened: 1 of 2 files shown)


mikemorris (approved):

mikemorris:
We should probably decide at some point if we want to follow a similar strategy as Consul core for older release branches, trying to stay on the latest patch of the Go minor version it started on, unless it becomes EoL while we're still maintaining and/or a CVE requires bumping to a newer Go minor version.

sarahethompson (approved):

PR #250, opened 2022-06-30
Use Terraform for conformance tests on EKS

nathancoleman:
Use Terraform instead of eksctl to manage our EKS cluster for conformance tests.
The main driver here is that eksctl frequently leaves orphaned resources that we have to manually clean up.
In addition, this gives us a workflow that's easily adaptable for GKE and AKS in the near future.

### How I've tested this PR:
See successful conformance EKS run on 0898eefcada0180e18cc4ee78ce4ed86f3824300

See above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -1,6 +1,9 @@
 name: Conformance (Main)
 
 on:
+  push:
+    branches: ["tf-conformance"]
+
   schedule:
     - cron:  '0 0 * * *'
 
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -1,9 +1,6 @@
 name: Conformance (Main)
 
 on:
-  push:
-    branches: [ 'tf-conformance']
-
   schedule:
     - cron:  '0 0 * * *'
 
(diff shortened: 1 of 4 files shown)


sarahalsmiller (approved):

.github/workflows/conformance.yml
```diff
@@ -1,6 +1,9 @@
 name: Conformance (Main)
 
 on:
+  push:
+    branches: ["tf-conformance"]
+
```
mikemorris, line 6: Guessing this was for testing and should be removed, maybe worth adding a separate GitHub label flag for `conformance/kind` vs `conformance/all` or something.

nathancoleman: It was for testing - will remove today.

`conformance/all` IMO implies your code is having conformance tests run against it, but that's not the case. This is because we don't build and publish images for a PR that can run on cloud providers. So really, all you can do with the cloud providers in a PR is test the setup code itself.

sarahalsmiller: We could always add something like `confromance/test-cloud-infrastructure`

PR #252, opened 2022-07-01
Add initial Dependabot configuration

nathancoleman:
Add initial Dependabot configuration to keep our dependencies up-to-date. This is a best practice and should lower the amount of CVEs we have to PR a remediation for if they're always updated.

### How I've tested this PR:
🤖 tests pass

See above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


mikemorris (approved):
Can always tweak this if it becomes annoying/noisy.

Issue #230, opened 2022-06-16
When deployment/service is deleted, after n hrs gateway listener is removed and never re-added


lkysow:
### Overview of the Issue

I have an HTTPRoute that points at two services. When I delete the services the listener is still configured on the gateway. After a certain time (12h?), Kubernetes triggers a re-resolve on the HTTPRoute and it's determined that the route is invalid since it targets non-running services, this causes the listener to be deleted.

If I then restart my services, the listener isn't added back. I have to restart the api gateway controller to get the listener back.

andrewstucki:
For some context, this happens because we're not properly re-reconciling routes when their `backendRef` services are deleted. When the service is deleted in Consul, all of its config entries get cleaned up, so there's still a listener, but it has no valid upstream rules (seems kind of odd that you can get Consul in this state, since you normally can't register a listener with 0 upstreams, though it seems you can register with 1 upstream, and then delete it).

After the 12h(?) period, Kubernetes triggers periodic reconciliation that sees we now have an invalid `backendRef` and removes the listener because it has no valid upstream rules (since Consul doesn't allow us to register a listener with no upstreams).

The restart of the services doesn't do anything for re-reconciling/updating the gateway config entries b/c of the lack of a service watch, but restarting the controller works because it causes the gateways/routes to re-reconcile yet again.

Fix should be to set up route reconciliations based off of upstream service lifecycle events -- we'll need to additionally add in some care to avoid race-conditions in our dirty checking code for our resolved gateway state.

Additionally, I have the feeling that a similar bug could happen in the presence of networking issues that cause a gateway to be de-registered as a service -- that the underlying config entries could be deleted automatically by Consul without us realizing we need to re-create them, so our fix should look into doing periodic re-syncs of all of our gateway state.

nathancoleman:
Thanks guys! I'll get started on a fix for this today. As discussed, it won't make our v0.3.0 release next week, but we can cut a patch release if necessary once the fix is merged.

nathancoleman:
Update here: I've got a working fix for this that I'll have up for code review on Monday

PR #278, opened 2022-07-11
Sync in-memory store w/ Consul at constant interval

nathancoleman:
This is longer-term followup based on https://github.com/hashicorp/consul-api-gateway/issues/230#issuecomment-1157976119

Trigger a sync for our in-memory store at a constant interval in the background

### How I've tested this PR:
_Still working on how to test this in a programmatic fasion_

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
@@ -294,7 +295,7 @@ func TestGatewayBasic(t *testing.T) {
 				}
 				service := services[0]
 				status := service.Checks.AggregatedStatus()
-				return status == "passing"
+				return status == api.HealthPassing
 			}, checkTimeout, checkInterval, "no healthy consul service found in the allotted time")
 
 			require.Eventually(t, gatewayStatusCheck(ctx, resources, gatewayName, namespace, conditionReady), checkTimeout, checkInterval, "no gateway found in the allotted time")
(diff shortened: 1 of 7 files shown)


internal/commands/server/k8s_e2e_test.go
```diff
@@ -299,7 +299,23 @@ func TestGatewayBasic(t *testing.T) {
...
+			// check for ingress gateway config entry
+			var entries []api.ConfigEntry
+			var err error
+			require.Eventually(t, func() bool {
+				entries, _, err = client.ConfigEntries().List(api.IngressGateway, &api.QueryOptions{Namespace: e2e.ConsulNamespace(ctx)})
+				return err == nil && len(entries) == 1
+			}, 5*time.Minute, checkInterval, "config entry not created in allotted time")
+
+			// verify background re-sync of config entries after delete
+			_, err = client.ConfigEntries().Delete(api.IngressGateway, entries[0].GetName(), &api.WriteOptions{Namespace: e2e.ConsulNamespace(ctx)})
```
andrewstucki, line 311: this is probably good enough, but if you want to exercise a bit more of the re-sync you could try and delete the gateway service from Consul directly. Deleting the service should remove all of its config entries as well, and we should have the service get re-registered via the registration goroutine in the deployment and the config entries subsequently re-created when the service appears again in the catalog.

nathancoleman: This sounds intriguing! Will dig into this today

internal/store/memory/store.go
```diff
@@ -2,20 +2,22 @@ package memory
...
 	"github.com/hashicorp/consul-api-gateway/internal/core"
 	"github.com/hashicorp/consul-api-gateway/internal/metrics"
 	"github.com/hashicorp/consul-api-gateway/internal/store"
-	"github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-multierror"
 )
 
 var (
-	ErrCannotBindListener = errors.New("cannot bind listener")
+	consulSyncInterval = 60 * time.Second
```
andrewstucki, line 19: we may want to consider moving this into a variable for command-line override eventually -- there should only be a few API calls per-gateway, but I would imagine that for some with large numbers of gateways deployed (not the normal case), this _could_ cause problems down the line if it's too aggressive. That said, 60s seems like it's fairly safe in like 99.9% of use-cases (i.e. non-overloaded Consul server or low number of gateways).

andrewstucki (approved):

internal/commands/server/k8s_e2e_test.go
nathancoleman: After some digging, I discovered that deregistering the service from Consul does not remove config entries and intentions like we thought. Given that, I added additional assertions for both services and config entries so the tests are now more robust.

PR #279, opened 2022-07-12
Consume gateway-api v0.5.0-rc2

nathancoleman:
Consume RC2 of gateway-api v0.5.0 as this is likely to be the GA release later this week. We at least want to make sure our tests all pass with it.

Additionally, this fixes a panic when using `ReferencePolicy` instead of `ReferenceGrant` due to the inclusion of https://github.com/kubernetes-sigs/gateway-api/pull/1239. This has been inhibiting my own testing of other changes.

### How I've tested this PR:
🤖 tests pass

See above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


sarahalsmiller (approved):

config/crd/kustomization.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.5.0-rc1
-- https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.5.0-rc1/config/crd/experimental/gateway.networking.k8s.io_referencepolicies.yaml
```
mikemorris, line 6: This had been necessary to install the ReferencePolicy CRD by default, as it had seemed to be excluded when only installing the experimental channel defaults due to being marked as deprecated - is that no longer the case?

nathancoleman: @mikemorris https://github.com/kubernetes-sigs/gateway-api/pull/1216 restored the `ReferencePolicy` CRD to the experimental channel

PR #289, opened 2022-07-14
Use pattern matching for release_branches

nathancoleman:
Pattern matching was recently added to our release tooling so that we no longer have to explicitly list every branch that should trigger release workflows. This simplifies our release preparation because we don't have to programmatically go in and add a new entry to `release_branches`.

### How I've tested this PR:
Confirmed that patterns in `release_branches` match the patterns used in our [build workflow](https://github.com/hashicorp/consul-api-gateway/blob/4d9bb77e40f3488e127f33b433dfa435bfa6d1f6/.github/workflows/build.yml#L6-L7). The contract for pattern matching in our release tooling is the same as GitHub's pattern matching.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -66,17 +66,6 @@ jobs:
         with:
           go-version: 1.18
 
-      - name: Update release_branches config
-        # TODO Make more robust in the future, could currently be a major or minor release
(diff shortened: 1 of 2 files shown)


sarahalsmiller (approved):
Thanks for being so quick on the draw

sarahethompson (approved):

PR #218, opened 2022-06-09
conformance: update hashicorp fork to skip failing tests

mikemorris:
### Changes proposed in this PR:

Update the conformance testing workflow to checkout an updated [branch](https://github.com/kubernetes-sigs/gateway-api/compare/master...hashicorp:conformance/v0.5.0-skipped-tests) on the hashicorp/gateway-api fork with a few failing tests skipped.

The longer-term plan will be to integrate the conformance suite directly as in #205 rather than relying on a fork, but we're holding off on that until gateway-api tags a v0.5.0 release.

### How I've tested this PR:

Observed the conformance tests are now passing.

### How I expect reviewers to test this PR:

Verify that this change is an adequate short-term resolution.

### Checklist:
- [ ] ~~Tests added~~
- [ ] ~~CHANGELOG entry added~~
  > Run `make changelog-entry` for guidance in authoring a changelog entry, and
  > commit the resulting file, which should have a name matching your PR number.
  > Entries should use imperative present tense (e.g. Add support for...)

PR #290, opened 2022-07-15
Implement InvalidKind reason for unknown/unsupported backendRef kind

nathancoleman:
The contract for `HTTPBackendRef` has changed to include a new reason, `InvalidKind`, for the `ResolvedRefs` condition when an unknown or unsupported backendRef kind is used ([docs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPBackendRef)). This change set implements that reason.

### How I've tested this PR:
Created HTTPRoute w/ unknown/unsupported kind for `backendRef`, verifying that the new `InvalidKind` reason is assigned to the status.
<details>
<summary>Example</summary>

By making a small change to the `backendRef` on this HTTPRoute from the Learn guide:
```yaml
kind: HTTPRoute
metadata:
  name: example-route-2
  namespace: consul
spec:
  parentRefs:
  - name: api-gateway
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - kind: Nonsense
      name: nginx
      namespace: consul
      port: 80
```
</details>

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/290.txt
+++ b/.changelog/290.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+Assign InvalidKind reason to ResolvedRefs condition on routes where the backend reference is an unknown or unsupported kind
+```
--- a/internal/k8s/reconciler/config/statuses.yaml
+++ b/internal/k8s/reconciler/config/statuses.yaml
@@ -154,6 +154,9 @@
             This reason is used with the “ResolvedRefs” condition when one of the Listener’s Routes has a BackendRef
             to an object in another namespace, where the object in the other namespace does not have a ReferenceGrant
             explicitly allowing the reference.
+        - name: InvalidKind
+          description: >
+            This reason is used when a Route references a backend with an unsupported group or kind.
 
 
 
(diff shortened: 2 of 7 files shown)


nathancoleman:
Debated adding e2e coverage for this, but it feels sorta like we're just duplicating upstream conformance tests. Thoughts?

sarahalsmiller:
> Debated adding e2e coverage for this, but it feels sorta like we're just duplicating upstream conformance tests. Thoughts?

If it's already covered in the upstream test, I don't see a reason to add an additional test.

mikemorris (approved):
Skimmed the upstream changes too quickly - per discussion in Gateway API Slack it appears there's not actually a conflict, as `InvalidKind` is to be used only for an invalid `BackendRef` kind, whereas `NotAllowedByListeners` would be used on when the _route_ kind (e.g. HTTPRoute) is not allowed by the listener `AllowedRoutes` config.

Separately from this change, we should likely open a draft PR updating the branch target for our conformance test fork (similar to https://github.com/hashicorp/consul-api-gateway/pull/218) to check if the new tests added upstream in v0.5.0-rc2 are passing, or need to be skipped due to other issues we don't yet support (partial acceptance of routes, specific expected response codes).

PR #291, opened 2022-07-15
Implement BackendNotFound reason for backendRef that does not exist

nathancoleman:
> Will want to review + merge #290 first which this PR is based on

The contract for `HTTPBackendRef` has changed to include a new reason, `BackendNotFound`, for the `ResolvedRefs` condition when a known backendRef kind is used but the backend does not exist ([docs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPBackendRef)). This change set implements that reason.

### How I've tested this PR:
Created HTTPRoute w/ non-existent `Service` for `backendRef`, verifying that the new `BackendNotFound` reason is assigned to the status.
<details>
<summary>Example</summary>

By making a small change to the `backendRef` on this HTTPRoute from the Learn guide:
```yaml
kind: HTTPRoute
metadata:
  name: example-route-2
  namespace: consul
spec:
  parentRefs:
  - name: api-gateway
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - kind: Service
      name: nginx-not-here
      namespace: consul
      port: 80
```
</details>
### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/service/resolver.go
+++ b/internal/k8s/service/resolver.go
@@ -28,6 +28,7 @@ type ServiceResolutionErrorType int
 
 const (
 	K8sServiceResolutionErrorType ServiceResolutionErrorType = iota
+	BackendNotFoundErrorType
 	ConsulServiceResolutionErrorType
 	GenericResolutionErrorType
 	InvalidKindErrorType
@@ -51,6 +52,10 @@ func NewResolutionError(inner string) ResolutionError {
 
 func NewK8sResolutionError(inner string) ResolutionError {
 	return ResolutionError{inner, K8sServiceResolutionErrorType}
+}
+
+func NewBackendNotFoundError(inner string) ResolutionError {
+	return ResolutionError{inner, BackendNotFoundErrorType}
 }
 
 func NewConsulResolutionError(inner string) ResolutionError {
@@ -237,7 +242,7 @@ func (r *backendResolver) consulServiceForK8SService(ctx context.Context, namesp
 	}
 	if service == nil {
 		r.logger.Warn("kubernetes service not found")
-		return nil, NewK8sResolutionError(fmt.Sprintf("service %s not found", namespacedName))
+		return nil, NewBackendNotFoundError(fmt.Sprintf("service %s not found", namespacedName))
 	}
 
 	// we do an inner retry since consul may take some time to sync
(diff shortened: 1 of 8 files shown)


mikemorris (changes requested):
~~LGTM to merge _before_ #290, as it's not impacted by the potential upstream logic conflict described in https://github.com/hashicorp/consul-api-gateway/pull/290#pullrequestreview-1043784170~~

I _think_ the usage of `NewConsulResolutionError` at https://github.com/hashicorp/consul-api-gateway/blob/f19a977cc0f66d9a88e0ec48860440e3b6c92e3c/internal/k8s/service/resolver.go#L257 could be updated to instead return `NewBackendNotFoundError` too.

I was curious if there are any remaining valid uses of `NewConsulResolutionError`, or could if it be removed now, but I'm not sure if https://github.com/hashicorp/consul-api-gateway/blob/f19a977cc0f66d9a88e0ec48860440e3b6c92e3c/internal/k8s/service/resolver.go#L280-L283 sufficiently maps to `NewBackendNotFoundError` to remove it entirely? Maybe?

~~Holding at least until #290 merges and this gets rebased for CI tests actually run.~~

Looks like there's a failing e2e test.

internal/k8s/service/resolver.go
```diff
@@ -385,7 +387,7 @@ func (r *backendResolver) findCatalogService(service *apigwv1alpha1.MeshService)
 		return nil, err
 	}
 	if len(services) == 0 {
-		return nil, NewConsulResolutionError(fmt.Sprintf("consul service (%s, %s) not found", consulNamespace, consulName))
+		return nil, NewBackendNotFoundError(fmt.Sprintf("consul service (%s, %s) not found", consulNamespace, consulName))
```
mikemorris, line 390: Are these other uses of `NewConsulResolutionError`, or could it be removed now?

andrewstucki:
@mikemorris yeah, I think that the line you commented here:

> I was curious if there are any remaining valid uses of NewConsulResolutionError, or could if it be removed now, but I'm not sure if ... sufficiently maps to NewBackendNotFoundError to remove it entirely? Maybe?

is different enough from not being able to find a backend -- i.e. it found one, but problematically it found more than one -- that it probably should be kept its own error type.

mikemorris (approved):

PR #280, opened 2022-07-12
Restore inclusion of ReferencePolicy in Grant lookup

nathancoleman:
### Changes proposed in this PR:
The ReferencePolicy needs to be included for backward compatibility; however, it couldn't be included until v0.5.0-rc2 of kubernetes-sigs/gateway-api which restored the ReferencePolicy CRD to the default set and fixed a listing bug.

v0.5.0-rc2 was consumed in #279, but I wasn't aware of this related TODO until I attempted the [Learn guide](https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway) which still uses the deprecated `ReferencePolicy` instead of the new `ReferenceGrant`.

### How I've tested this PR:
- [x] 🤖 tests pass
- [x] [Learn guide](https://learn.hashicorp.com/tutorials/consul/kubernetes-api-gateway) works

### How I expect reviewers to test this PR:
See above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
  > Run `make changelog-entry` for guidance in authoring a changelog entry, and
  > commit the resulting file, which should have a name matching your PR number.
  > Entries should use imperative present tense (e.g. Add support for...)

PR #293, opened 2022-07-19
referencepolicy: finish backwards compatibility support and tests

mikemorris:

- [x] Simplify ReferencePolicy to ReferenceGrant conversion logic in gateway client
  - To match https://github.com/hashicorp/consul-api-gateway/blob/main/internal/k8s/controllers/http_route_controller.go#L141, possible since ReferencePolicy now uses ReferenceGrantSpec internally as the type of its `Spec` field.
- [x] Fixup comment to properly refer to ReferenceGrant instead of ReferencePolicy
- [x] Add tests for Gateway certificateRefs ReferenceGrant, and tests for ReferencePolicy backwards compatibility
- [x] Add e2e tests for ReferencePolicy backwards compatibility

### How I've tested this PR:

CI tests should still be passing after changes.

Verify changes make sense and sufficiently cover support.

### Checklist:
- [x] Tests added

### DIFF
--- a/internal/k8s/gatewayclient/gatewayclient.go
+++ b/internal/k8s/gatewayclient/gatewayclient.go
@@ -545,22 +545,7 @@ func (g *gatewayClient) GetReferenceGrantsInNamespace(ctx context.Context, names
 		return nil, NewK8sError(err)
 	}
 	for _, refPolicy := range refPolicyList.Items {
-		refGrant := gwv1alpha2.ReferenceGrant{}
-		for _, refPolicyFrom := range refPolicy.Spec.From {
-			refGrant.Spec.From = append(refGrant.Spec.From, gwv1alpha2.ReferenceGrantFrom{
-				Group:     refPolicyFrom.Group,
-				Kind:      refPolicyFrom.Kind,
-				Namespace: refPolicyFrom.Namespace,
-			})
-		}
-		for _, refPolicyTo := range refPolicy.Spec.To {
-			refGrant.Spec.To = append(refGrant.Spec.To, gwv1alpha2.ReferenceGrantTo{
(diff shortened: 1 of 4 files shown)


internal/k8s/gatewayclient/gatewayclient.go
```diff
@@ -545,22 +545,7 @@ func (g *gatewayClient) GetReferenceGrantsInNamespace(ctx context.Context, names
...
-		}
-		for _, refPolicyTo := range refPolicy.Spec.To {
-			refGrant.Spec.To = append(refGrant.Spec.To, gwv1alpha2.ReferenceGrantTo{
-				Group: refPolicyTo.Group,
-				Kind:  refPolicyTo.Kind,
-				Name:  refPolicyTo.Name,
-			})
-		}
-		refGrants = append(refGrants, refGrant)
+		refGrants = append(refGrants, gwv1alpha2.ReferenceGrant{Spec: refPolicy.Spec})
```
mikemorris, line 548: e2e test for ReferencePolicy backwards compatibility fails as expected when this line is commented out, so the test coverage gap that existed when https://github.com/hashicorp/consul-api-gateway/pull/280 was discovered is covered now.

sarahalsmiller (approved):
Will there be a point in time when referencepolicy is expected to fail, or will it keep working with the warning indefinitely?

nathancoleman (approved):

mikemorris:
> Will there be a point in time when referencepolicy is expected to fail, or will it keep working with the warning indefinitely?

I believe the upstream plan is to remove it in the next minor (v0.6.0) Gateway API release, so we would likely remove support when upgrading to that - could be several months out though.

PR #295, opened 2022-07-28
Implement cert+key retrieval using Vault's PKI API

nathancoleman:
As we ramp up to supporting deployment on VMs, we need an alternative to K8s Secrets as a source for TLS certificate + private key pairs.

Notably, this PR only includes the actual retrieval of a cert+key using [Vault's PKI API](https://www.vaultproject.io/api-docs/secret/pki). There is no way to actually hit this codepath yet and no user-facing change.

Future PRs will introduce watches for these certs as well as a method of configuring a Gateway to use them.

### How I've tested this PR:
- Run Vault locally (I used [HashiQube](https://github.com/servian/hashiqube))
- Follow the first three steps of the tutorial for setting up PKI [here](https://developer.hashicorp.com/vault/tutorials/secrets-management/pki-engine#step-1-generate-root-ca)
- After reaching the end of Step 3, you can run a script such as the one below, using the new `vault.SecretClient` to successfully generate a cert and key using the intermediate CA and role that you created.

<details>
<summary>Example</summary>

```go
package main

import (
	"context"
	"fmt"

	"github.com/hashicorp/go-hclog"

	"github.com/hashicorp/consul-api-gateway/internal/vault"
)

const (
	issue      = "example-dot-com"
	pkiPath    = "pki_int"
	secretName = "vault://test.example.com?ttl=24h"
)

func main() {
	client, err := vault.NewSecretClient(hclog.New(hclog.DefaultOptions), pkiPath, issue)
	if err != nil {
		panic("failed to create vault.SecretClient: " + err.Error())
	}

	secret, expires, err := client.FetchSecret(context.Background(), secretName)
	if err != nil {
		panic("failed to fetch secret: " + err.Error())
	}

	fmt.Println(fmt.Sprintf("Fetched secret %q\n\tExpires: %s\n\tSecret: %+v", secretName, expires, secret))
}
```
</details>

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/commands/server/server.go
```diff
@@ -127,3 +126,23 @@ func RunServer(config ServerConfig) int {
...
+		return nil, err
+	}
+	secretClient.Register(utils.K8sSecretScheme, k8sSecretClient)
+
+	vaultSecretClient, err := vault.NewSecretClient(config.Logger.Named("vault-cert-fetcher"), "pki", "TODO")
+	if err != nil {
+		config.Logger.Error("error initializing the vault secret fetcher", "error", err)
+		return nil, err
+	}
+	secretClient.Register(vault.SecretScheme, vaultSecretClient)
```
mikemorris, line 145: 👍 This looks cleaner than the prior `AddToMultiClient` method that mutated the argument passed in.

internal/vault/certificates.go
```diff
@@ -26,13 +26,18 @@ const (
 	defaultIssuer = "default"
 )
 
+//go:generate mockgen -source ./certificates.go -destination ./mocks/certificates.go -package mocks LogicalClient
+type LogicalClient interface {
```
mikemorris, line 30: I think the `Logical` naming is a bit awkward/confusing, particularly the `.Logical()` getter, could this be `VaultClient` or something instead?

mikemorris (changes requested):
Mostly LGTM aside from a naming concern, this is a case where I do wish we had some test coverage tooling but looks adequate for initial implementation to me.

internal/vault/certificates.go
andrewstucki: I'm ok with the `LogicalClient`, specifically because it makes it awkward having to nest two interfaces for a call to `Logical`/`WriteWithContext`

andrewstucki (approved):
I'm cool with this as an initial pass with the mock-based testing, just wondering though, is there a plan to add integration-level testing with a real Vault installation as a follow-up?

nathancoleman:
@andrewstucki Absolutely! We're planning to add Vault to our automated testing stack in a separate PR as part of this effort, it's just a whole chunk of work on its own to stand it up, configure PKI, and then expose this codepath in some way that's testable from the "outside world".

internal/vault/certificates.go
nathancoleman: I initially went with VaultClient, but it felt kinda gross since you either nest two interface like Andrew said, or you wind up labeling a very specific set of the Vault API/SDK as `VaultClient` and assuming we'll never need the rest of the API/SDK outside of `vault.NewClient(...).Logical()`. I'm not super opinionated either way though 🤷

andrewstucki: also the `vault.VaultClient` interface would stutter more than I like 😅

mikemorris (approved):

PR #302, opened 2022-08-09
TLS Watch and Rotate

sarahalsmiller:
- To account for the fact that certs can technically be rotated at any time, after a set interval force a repull of all certificates being watched. 

### How I've tested this PR:
- Tests pass
- Locally using Hashiqube to run vault and kind to run api gateway

- Tests pass

### Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/envoy/secrets.go
+++ b/internal/envoy/secrets.go
@@ -311,7 +311,7 @@ func (s *secretManager) Manage(ctx context.Context, forceInterval time.Duration)
 		case <-time.After(s.loopTimeout):
 			force := false
 			now := time.Now()
-			if now.Sub(lastForce) > forceInterval {
+			if now.Sub(lastForce) >= forceInterval {
 				force = true
 				lastForce = now
 			}
--- a/internal/envoy/secrets.go
+++ b/internal/envoy/secrets.go
@@ -307,7 +307,7 @@ func (s *secretManager) Manage(ctx context.Context, forceInterval time.Duration)
 	for {
 		select {
 		case <-time.After(s.loopTimeout):
-			s.manage(ctx, force)
+			s.manage(ctx, false)
 		case <-time.After(forceInterval):
 			s.manage(ctx, true)
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


internal/envoy/secrets.go
```diff
@@ -301,29 +301,38 @@ func (s *secretManager) unwatch(names []string, node string) error {
...
+			now := time.Now()
+			if now.Sub(lastForce) >= forceInterval {
+				force = true
+				lastForce = now
+			}
+			s.manage(ctx, force)
 		case <-ctx.Done():
 			// we finished the context, just return
 			return
 		}
```
nathancoleman, line 322: I think you could simplify a bit at the risk of pulling twice rapidly due to the offset between the two being random. I don't imagine that would be an actual problem operationally though

```go
select {
case <-time.After(s.loopTimeout):
    s.manage(ctx, false)
case <-time.After(forceInterval):
    s.manage(ctx, true)
case <-ctx.Done():
    // we finished the context, just return
    return
}
```

internal/envoy/secrets.go
```diff
@@ -301,29 +301,38 @@ func (s *secretManager) unwatch(names []string, node string) error {
 }
 
 // Manage is used for re-fetching expiring TLS certificates and updating them
```
nathancoleman, line 303: Some extra context about what this function is doing in the docstring here would be helpful I think. It's not immediately obvious that we're checking on an interval to see if our cached cert has expired and also force-refreshing our cached cert on a separate interval.

internal/envoy/secrets.go
```diff
@@ -300,30 +300,33 @@ func (s *secretManager) unwatch(names []string, node string) error {
 	return nil
 }
 
-// Manage is used for re-fetching expiring TLS certificates and updating them
-func (s *secretManager) Manage(ctx context.Context) {
+// Manage is used for re-fetching expiring TLS certificates and updating them, additionally will force re-fetch all certificates every force intercal
```
nathancoleman, line 303:
```suggestion
// Manage is used for re-fetching expiring TLS certificates and updating them, additionally will force re-fetch all certificates every force interval
```

nathancoleman (approved):
This is a great starting point that we can iterate on further if needed for our v0.5.0 release.

It will result in refetching TLS certs that are stored in K8s Secrets today every 30 minutes, so we'll want to make sure to exercise that.

PR #308, opened 2022-08-15
Fix intention syncing CAS index

andrewstucki:

Our CAS operation for updating intentions was broken because it was using the wrong config entry index. The result is that if multiple gateways bound to a single service then intentions for the second gateway weren't being properly created. This changes the sync code to use the proper index.

### How I've tested this PR:

Manually and updated tests.

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/308.txt
+++ b/.changelog/308.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix intentions syncing for multiple gateways bound to a single route.
+```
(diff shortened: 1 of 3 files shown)


nathancoleman (approved):

nathancoleman:
Tested by setting a breakpoint on the modified assertion in `k8s_e2e_test.go`. Without this change, only one intention is created, and the second fails. With this change, both intentions are successfully created.

PR #313, opened 2022-08-16
Remove logic for handling unsupported UDPRoute + TLSRoute

nathancoleman:
In working to get the store refactor merged, this was one chunk of changes that I noticed could be broken out of #160 to make reviews less taxing. This simply removes logic that exists today for handling `UDPRoute` and `TLSRoute` which we do not support.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/reconciler/listener_test.go
```diff
@@ -537,7 +537,7 @@ func TestListenerCanBind_RouteKind(t *testing.T) {
 	routeMeta.SetGroupVersionKind(schema.GroupVersionKind{
 		Group:   gwv1alpha2.GroupVersion.Group,
 		Version: gwv1alpha2.GroupVersion.Version,
-		Kind:    "UDPRoute",
+		Kind:    "TCPRoute",
```
nathancoleman, line 540: Simply needed a substitue for the test, and the `Kind` here only needed to be incompatible with the protocol below, `http`

internal/k8s/reconciler/route_test.go
```diff
@@ -150,8 +134,8 @@ func TestRouteMergedStatusAndBinding(t *testing.T) {
 	}, K8sGatewayConfig{
 		Logger: hclog.NewNullLogger(),
 	})
-	inner := &gwv1alpha2.TLSRoute{
-		Spec: gwv1alpha2.TLSRouteSpec{
+	inner := &gwv1alpha2.HTTPRoute{
```
nathancoleman, line 137: Kind here doesn't matter, just needed _a route_

nathancoleman (commented):
Personal review

mikemorris (approved):
LGTM, hopefully when we actually introduce support for these we can reduce some of the common xRoute boilerplate to be less duplicative.

PR #314, opened 2022-08-17
Split error + status generation into separate subpackages of reconciler

nathancoleman:
In working to get the store refactor merged, this was one chunk of changes that I noticed could be broken out of https://github.com/hashicorp/consul-api-gateway/pull/160 to make reviews less taxing. This simply moves the error and status definitions into purpose-specific packages and splits the single `generator.go` file into the specific parts that belong in each package.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


andrewstucki (approved):
Looks pretty much just lifted from the original PR and low impact, so I'm 👍

PR #317, opened 2022-08-18
Refactor K8sGateway state into serializable struct

nathancoleman:
In working to get the store refactor merged, this was one chunk of changes that I noticed could be broken out of https://github.com/hashicorp/consul-api-gateway/pull/160 to make reviews less taxing. This simply moves the state of the `K8sGateway` into a separate struct that can be serialized (important in the future).

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/common/utils.go
+++ b/internal/k8s/reconciler/common/utils.go
@@ -24,12 +24,13 @@ var (
 	}
 )
 
-// SupportedKindsFor --
+// SupportedKindsFor returns the list of xRoute Kinds that support a given protocol
 func SupportedKindsFor(protocol gwv1beta1.ProtocolType) []gwv1beta1.RouteGroupKind {
 	return supportedKindsForProtocol[protocol]
 }
 
-// AsJSON --
+// AsJSON serializes a given item into a JSON string. The item is assumed to be
+// JSON-serializable as this function will panic otherwise.
 func AsJSON(item interface{}) string {
 	data, err := json.Marshal(item)
 	if err != nil {
@@ -42,7 +43,8 @@ func AsJSON(item interface{}) string {
 	return string(data)
 }
 
-// ParseParent --
+// ParseParent deserializes a JSON string into a gwv1alpha2.ParentReference. The string
+// is assumed to be a valid JSON representation as this function will panic otherwise.
 func ParseParent(stringified string) gwv1alpha2.ParentReference {
 	var ref gwv1alpha2.ParentReference
 	if err := json.Unmarshal([]byte(stringified), &ref); err != nil {
--- a/internal/k8s/reconciler/common/utils.go
+++ b/internal/k8s/reconciler/common/utils.go
@@ -1,6 +1,9 @@
 package common
 
 import (
+	"encoding/json"
+
+	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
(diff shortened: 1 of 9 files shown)


internal/k8s/reconciler/common/utils.go
```diff
@@ -0,0 +1,56 @@
...
+			Kind:  "HTTPRoute",
+		}},
+		gwv1beta1.TCPProtocolType: {{
+			Group: (*gwv1beta1.Group)(&gwv1beta1.GroupVersion.Group),
+			Kind:  "TCPRoute",
+		}},
+	}
+)
+
+// SupportedKindsFor --
```
andrewstucki, line 27: While we're at it, mind cleaning up these docstrings? 😅

internal/k8s/reconciler/state/gateway.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return gwv1beta1.GatewayStatus{
+		Addresses:  addresses,
+		Conditions: conditions,
+		Listeners:  listenerStatuses,
+	}
+}
+
+// ListenerState holds ephemeral state for listeners
+type ListenerState struct {
```
andrewstucki, line 72: I don't think that this is currently in-use? IIRC the original PR did some modification of the K8sListener struct to try and tie the `ListenerState` onto the `K8sListener` before removing the `K8sListener` in the follow-ups. Without this in use I think there might be issues trying to call things like the `GatewayState.GetStatus` which calculates statuses based off of this state.

andrewstucki (changes requested):
I think for this to be functional it's going to need to tie the `ListenerState` onto the `K8sListener` structs during validation, no?

internal/k8s/reconciler/state/gateway.go
nathancoleman: Good catch!

This struct currently gets consumed up in `GatewayState` and now, with commits since you reviewed, is initialized on the `K8sListener` as part of `NewK8sGateway`. I believe this leaves the only difference from the [original refactor](https://github.com/hashicorp/consul-api-gateway/pull/160/files#diff-53c61379ab3051a987420aef6574a34645539c538cd1250d49a93b584667724b) being that this logic moves into the `Factory`. That change is incoming on the PR after this one.

Lemme know if ya think I've missed something

internal/k8s/reconciler/gateway.go
```diff
@@ -49,21 +49,22 @@ type K8sGatewayConfig struct {
 }
 
 func NewK8sGateway(gateway *gwv1beta1.Gateway, config K8sGatewayConfig) *K8sGateway {
+	// FUTURE (nathancoleman) See if we can avoid setting ConsulNamespace out of band
+	gwState := state.InitialGatewayState(gateway)
+	gwState.ConsulNamespace = config.ConsulNamespace
```
andrewstucki, line 54: The `ConsulNamespace` mapping code can probably happen as part of the validators if we do go ahead with that piece of the refactor later (I don't recall if it got mapped there or not in the initial refactor) -- but since it's a product of the gateway configuration (i.e. the namespace the gateway is created in) and some additional global configuration (i.e. the mapping configuration) -- it'd make sense to me to handle it as part of the initial "validation" step that calculates whatever overall state/status changes that will eventually get applied to the store/Consul/K8s statuses.

andrewstucki (approved):

internal/k8s/reconciler/state/gateway.go
andrewstucki: So, just be aware that by the time the final PR rolled around, the factory actually got completely dropped since it was no longer necessary. If you have any ideas to avoid the intermediate introduction of the factories while keeping this fairly clean/in test-passable chunks, feel free to drop their introduction. Otherwise, I'm fine with going the same route as before just so we don't have to think too hard about this 😅

PR #320, opened 2022-08-19
Refactor K8sRoute state into serializable struct

nathancoleman:
In working to get the store refactor merged, this was one chunk of changes that I noticed could be broken out of https://github.com/hashicorp/consul-api-gateway/pull/160 to make reviews less taxing. This simply moves the state of the `K8sRoute` into a separate struct that can be serialized (important in the future).

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


andrewstucki:
So, this looks like the introduction of the `RouteState` code (similar in concept to the `GatewayState` code added previously) and the light touches on the comparison functions to strip out the logic from `K8sRoute`, correct?

nathancoleman:
@andrewstucki exactly 👍

internal/k8s/reconciler/route.go
```diff
@@ -353,7 +195,7 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 					nsName := getNamespacedName(ref.Name, ref.Namespace, route.Namespace)
 					msg := fmt.Sprintf("Cross-namespace routing not allowed without matching ReferenceGrant for Service %q", nsName)
 					r.logger.Warn("Cross-namespace routing not allowed without matching ReferenceGrant", "refName", nsName.Name, "refNamespace", nsName.Namespace)
-					r.resolutionErrors.Add(service.NewRefNotPermittedError(msg))
+					r.RouteState.ResolutionErrors.Add(service.NewRefNotPermittedError(msg))
```
andrewstucki, line 198: it'll be so nice when we get to the validator and other service structs later -- these multi-chained function calls are gross 😢

andrewstucki (approved):
Overall this seems like it's pretty much the changes introduced in the original PR -- aside from breaking the changes apart into this smaller chunk, was there any functional logic issues that needed to be updated? I didn't see any, but I know we'll run across some harder merge conflicts later with updated logic.

nathancoleman:
@andrewstucki so far the difficult part is just working through things that have moved in the refactor but were also modified or had new functionality added in `main`, no big functional changes to work through though.

PR #321, opened 2022-08-19
Introduce Factory for creating Gateways + Routes

nathancoleman:
In working to get the store refactor merged, this was one chunk of changes that I noticed could be broken out of https://github.com/hashicorp/consul-api-gateway/pull/160 to make reviews less taxing. This introduces a factory for creating Gateways and Routes.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/route_test.go
+++ b/internal/k8s/reconciler/route_test.go
@@ -31,13 +31,13 @@ func TestRouteID(t *testing.T) {
 		Namespace: "namespace",
 	}
 
-	require.Equal(t, "http-namespace/name", NewK8sRoute(&gwv1alpha2.HTTPRoute{
+	require.Equal(t, "http-namespace/name", newK8sRoute(&gwv1alpha2.HTTPRoute{
 		ObjectMeta: meta,
 	}, config).ID())
-	require.Equal(t, "tcp-namespace/name", NewK8sRoute(&gwv1alpha2.TCPRoute{
+	require.Equal(t, "tcp-namespace/name", newK8sRoute(&gwv1alpha2.TCPRoute{
 		ObjectMeta: meta,
 	}, config).ID())
-	require.Equal(t, "", NewK8sRoute(&core.Pod{
+	require.Equal(t, "", newK8sRoute(&core.Pod{
 		ObjectMeta: meta,
 	}, config).ID())
 }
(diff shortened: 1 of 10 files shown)


internal/k8s/service/resolver.go
```diff
@@ -182,31 +182,28 @@ func (r *ResolvedReference) Item() client.Object {
...
-	logger    hclog.Logger
-	mapper    common.ConsulNamespaceMapper
+	client gatewayclient.Client
+	consul *api.Client
+	logger hclog.Logger
+	mapper common.ConsulNamespaceMapper
 }
 
-func NewBackendResolver(logger hclog.Logger, namespace string, mapper common.ConsulNamespaceMapper, client gatewayclient.Client, consul *api.Client) BackendResolver {
+func NewBackendResolver(logger hclog.Logger, mapper common.ConsulNamespaceMapper, client gatewayclient.Client, consul *api.Client) *backendResolver {
```
nathancoleman, line 195: Removing the `namespace` from the resolver itself and instead providing it as an arg to `Resolve()` lets us re-use one resolver for backends across different namespaces 💯

internal/k8s/reconciler/route_test.go
```diff
@@ -173,8 +173,6 @@ func TestRouteValidate(t *testing.T) {
 		Reference: &service.BackendReference{},
 	}
 
-	resolver.EXPECT().Resolve(gomock.Any(), reference).Return(resolved, nil)
-
 	route := NewK8sRoute(&gwv1alpha2.HTTPRoute{
```
andrewstucki, line 176: Not sure if I missed some stuff like this in the initial PR or if it's not that big of a deal because it eventually gets dropped anyway -- but thinking we should drop the public initializer functions in this PR if we're introducing the factories.

andrewstucki (changes requested):
If we can drop the initializer functions that are outside of the new `Factory` struct, think this should be good to go.

Is this more or less the last PR for the 1/3 part of the refactor?

internal/k8s/reconciler/route_test.go
nathancoleman: @andrewstucki Thoughts on just making them private instead of public? In particular, `NewK8sGateway` has a good bit of "special" init logic for pulling listeners off of the kubernetes `Gateway` object. It feels better to me to have that live in `gateway.go` for example rather than `factory.go`.

andrewstucki: I'm fine with that for now, especially as this code is very much "in-flux" until we're closer to the full refactor merge, at which point we can deal with changing this up if necessary.

andrewstucki (approved):
This looks fine to merge for now. It should get dropped as part of a later PR if you follow the older refactor code.

We can deal with any changes to the initializers later if need be, don't want to stall making progress.

internal/k8s/reconciler/route_test.go
nathancoleman: Un-exported `NewK8sRoute` and `NewK8sGateway` 💯

PR #323, opened 2022-08-23
Extract validation logic for Gateway into GatewayValidator

nathancoleman:
This is a portion of the refactor changes from #163. The end result of this PR is that the logic for validating `Gateways` and `Listeners` is abstracted out into the `GatewayValidator`.

Instead of calling `K8sGateway.Validate()` - which in turn called `K8sListener.Validate()` - you call `GatewayValidator.Validate(v1beta1.Gateway)` which processes the Gateway and all Listeners and returns a `GatewayState`. This `GatewayState` is then paired with the `v1beta1.Gateway` to make up our internal model, `K8sGateway`.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/reconciler/manager.go
```diff
@@ -177,23 +181,29 @@ func (m *GatewayReconcileManager) UpsertGateway(ctx context.Context, g *gwv1beta
...
 	})
 
-	// Calling validate outside of the upsert process allows us to re-resolve any
-	// external references and set the statuses accordingly. Since we actually
-	// have other object updates triggering reconciliation loops, this is necessary
-	// prior to dirty-checking on upsert.
-	if err := gateway.Validate(ctx); err != nil {
-		return err
-	}
-
```
nathancoleman, line 196: The general flow now is that we:
1. Create a default/initial `GatewayState`
2. Validate the `Gateway` which updates the `GatewayState`
3. Create our internal model - `K8sGateway` - which pairs the derived `GatewayState` with the `Gateway`

internal/k8s/reconciler/utils.go
```diff
@@ -121,35 +121,6 @@ func routeAllowedForListenerNamespaces(ctx context.Context, gatewayNS string, al
 	return false, nil
 }
 
-// gatewayAllowedForSecretRef determines whether the gateway is allowed
-// for the secret either by being in the same namespace or by having
-// an applicable ReferenceGrant in the same namespace as the secret.
-func gatewayAllowedForSecretRef(ctx context.Context, gateway *gwv1beta1.Gateway, secretRef gwv1beta1.SecretObjectReference, c gatewayclient.Client) (bool, error) {
```
nathancoleman, line 127: This is only used during the validation process that determines what the `GatewayState` should be, so it's moved into the `validator` package

internal/k8s/reconciler/utils_test.go
```diff
@@ -357,115 +357,6 @@ func TestGatewayStatusEqual(t *testing.T) {
 	}))
 }
 
-func TestGatewayAllowedForSecretRef(t *testing.T) {
```
nathancoleman, line 360: This test has moved into `validator/utils_test.go`

internal/k8s/reconciler/validator/utils_test.go
```diff
@@ -0,0 +1,124 @@
+package validator
```
nathancoleman, line 1: This was moved over from `reconciler/utils_test.go`

internal/k8s/reconciler/validator/utils.go
```diff
@@ -0,0 +1,111 @@
+package validator
```
nathancoleman, line 1: This code was moved over from `reconciler/utils.go`

internal/k8s/reconciler/validator/gateway_test.go
```diff
@@ -0,0 +1,713 @@
+package validator
```
nathancoleman, line 1: These tests were moved over from `reconciler/gateway_test.go` and `reconciler/listener_test.go`

internal/k8s/reconciler/validator/gateway.go
```diff
@@ -0,0 +1,512 @@
+package validator
```
nathancoleman, line 1: This code was largely validation logic that was moved over from `reconciler/gateway.go` and `reconciler/listener.go`.

nathancoleman (commented):
Personal review

internal/k8s/reconciler/gateway.go
```diff
@@ -76,261 +69,6 @@ func newK8sGateway(gateway *gwv1beta1.Gateway, config K8sGatewayConfig) *K8sGate
 	}
 }
 
-func (g *K8sGateway) Validate(ctx context.Context) error {
```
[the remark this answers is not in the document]
nathancoleman: The big banner change for this PR is that this call to `Validate()` now just proxies through to the `GatewayValidator` below

internal/k8s/reconciler/gateway.go
```diff
@@ -77,258 +72,7 @@ func newK8sGateway(gateway *gwv1beta1.Gateway, config K8sGatewayConfig) *K8sGate
...
-}
-
-func (g *K8sGateway) validatePodStatusRunning(pod *corev1.Pod) {
-	for _, condition := range pod.Status.Conditions {
-		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
-			g.GatewayState.PodReady = true
-			return
-		}
-	}
+	return g.validator.Validate(ctx, g.GatewayState, g.Gateway, g.deployer.Service(g.config, g.Gateway))
```
andrewstucki, line 75: Wondering, did you keep the validator as part of the gateway to reduce the test changes that would need to be introduced to strip out `Validate` from the `K8sGateway` struct? I believe that the original PR had this call in the Manager? Is there a plan to move this there eventually?

andrewstucki (commented):
Generally looks more or less right, will take another pass, but have a question about overall usage of the validators.

internal/k8s/reconciler/gateway.go
nathancoleman: You are correct, and no reason we can't move it out to the manager in a future PR. This PR is already more chunky than I'd like though so tried to narrow as much as possible.

internal/k8s/reconciler/gateway.go
```diff
@@ -27,10 +22,11 @@ type K8sGateway struct {
...
 
-	logger   hclog.Logger
-	client   gatewayclient.Client
-	config   apigwv1alpha1.GatewayClassConfig
-	deployer *GatewayDeployer
+	logger    hclog.Logger
+	client    gatewayclient.Client
+	config    apigwv1alpha1.GatewayClassConfig
+	validator *validator.GatewayValidator
+	deployer  *GatewayDeployer
```
andrewstucki, line 29: Do you think you can add a `TODO` for removing these from the struct when we get to where this will be extracted into the manager? Just don't want to forget these. Same with the `client`/`logger` fields. Maybe we can just lump together the things to get dropped in subsequent PRs just so we know this is still in intermediate state.

internal/k8s/reconciler/listener.go
```diff
@@ -47,12 +48,23 @@ type K8sListenerConfig struct {
 	ConsulNamespace string
 	Logger          hclog.Logger
 	Client          gatewayclient.Client
+	State           *state.ListenerState
```
andrewstucki, line 51: I'm assuming this is unused? Looks like `ListenerState` is what you're using elsewhere.

internal/k8s/reconciler/listener.go
```diff
@@ -272,17 +118,17 @@ func (l *K8sListener) Config() store.ListenerConfig {
...
+	protocol, tls := utils.ProtocolToConsul(l.ListenerState.Protocol)
+	l.ListenerState.TLS.Enabled = tls
 
 	return store.ListenerConfig{
 		Name:     name,
 		Hostname: hostname,
 		Port:     int(l.listener.Port),
 		Protocol: protocol,
-		TLS:      l.tls,
+		TLS:      l.ListenerState.TLS,
```
andrewstucki, line 131: 👍

internal/k8s/reconciler/listener.go
```diff
@@ -368,12 +214,14 @@ func (l *K8sListener) canBind(ctx context.Context, ref gwv1alpha2.ParentReferenc
 	return false, nil
 }
 
-func (l *K8sListener) OnRouteAdded(_ store.Route) {
+func (l *K8sListener) OnRouteAdded(route store.Route) {
 	atomic.AddInt32(&l.routeCount, 1)
+	l.ListenerState.Routes[route.ID()] = *route.Resolve(l)
```
andrewstucki, line 219: 👍

andrewstucki (changes requested):
Mainly just drop the duplicate field reference, besides that pretty close.

internal/k8s/reconciler/listener.go
nathancoleman: This is `K8sListenerConfig.State` (arg to the constructor) vs. `K8sListener.ListenerState`

internal/k8s/reconciler/gateway.go
nathancoleman: I can. Noting that the end result for part 2 of the refactor still has `deployer` living on this struct so I'm assuming it's removed in refactor part 3. Will likely wait to move the validator until the same time.

internal/k8s/reconciler/listener.go
andrewstucki: So, it looks like it's not set at all in the constructor diff starting at line 57, so still thinking it's unused. Either way, if it isn't unused, these should be collapsed into a single field, since there should only ever be a single state associated with a listener:

```go
	lState := config.State
	if lState == nil {
		lState = &state.ListenerState{
			Name:     listener.Name,
			Protocol: listener.Protocol,
			Routes:   make(map[string]core.ResolvedRoute),
		}
	}

	return &K8sListener{
		ListenerState:   lState,
```

andrewstucki: oh gosh, nvm, the diff above was confusing me, disregard this, I see what you're saying

internal/k8s/reconciler/gateway.go
nathancoleman: Done

andrewstucki (approved):

PR #329, opened 2022-08-26
Use GatewayState for K8sGateway.Status()

nathancoleman:
The `GatewayState` should now be used for determining Gateway status, so just pipe the result back up to the caller instead of having duplicate logic.

### How I've tested this PR:
🤖  tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/reconciler/gateway.go
```diff
@@ -390,50 +390,7 @@ func (g *K8sGateway) ShouldBind(route store.Route) bool {
 }
 
 func (g *K8sGateway) Status() gwv1beta1.GatewayStatus {
-	listenerStatuses := []gwv1beta1.ListenerStatus{}
```
nathancoleman, line 393: The logic lives in `GatewayState` [here](https://github.com/hashicorp/consul-api-gateway/blob/eb308e0b33287eb7dbea7708cc659aaa24c2910c/internal/k8s/reconciler/state/gateway.go#L39-L69)

PR #330, opened 2022-08-29
Extract validation logic for Route into RouteValidator

nathancoleman:
Analogous to #323.

This is a portion of the refactor changes from #163. The end result of this PR is that the logic for validating `xRoutes` is abstracted out into the `RouteValidator`. The route then proxies through to `validator.Validate(...)` rather than containing all of the validation logic itself.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/k8s/reconciler/validator/route.go
```diff
@@ -0,0 +1,154 @@
...
+
+// all kubernetes routes implement the following two interfaces
+type Route interface {
+	client.Object
+	schema.ObjectKind
+}
+
+// RouteValidator is responsible for taking a provided v1alpha2.HTTPRoute or
+// v1alpha2.TCPRoute and deriving a state.RouteState from it. Ultimately, this
+// RouteState is what makes up the Status on the Kubernetes Route resource.
```
andrewstucki, line 26:
```suggestion
// RouteState is what makes up the Status on the Kubernetes Route resource
// and holds any intermediate state resolved from remote object references
// (i.e. resolved Consul services from `backendRefs` values).
```

andrewstucki (approved):
LGTM once tests run/pass.

internal/k8s/reconciler/validator/route.go
andrewstucki: Also, looks like we have a mismatch for `v1beta1.Gateway` in the above docstring v. `v1alpha2.HTTP/TCPRoute` here.

internal/k8s/reconciler/validator/utils.go
```diff
@@ -11,35 +11,6 @@ import (
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/gatewayclient"
 )
 
-// gatewayAllowedForSecretRef determines whether the gateway is allowed
-// for the secret either by being in the same namespace or by having
-// an applicable ReferenceGrant in the same namespace as the secret.
-func gatewayAllowedForSecretRef(ctx context.Context, gateway *gwv1beta1.Gateway, secretRef gwv1beta1.SecretObjectReference, c gatewayclient.Client) (bool, error) {
```
nathancoleman, line 17: Moved into `validator/gateway.go` to maintain symmetrical package structure

PR #332, opened 2022-08-30
Extract binding logic into Binder

nathancoleman:
This is a portion of the refactor changes from #163. The end result of this PR is that the logic for binding `xRoutes` to `Gateways` is abstracted out into `Binder`. You create a `Binder` using `binder.NewBinder( <Gateway> )` and then use it to bind routes by calling `<Binder>.Bind( <Route> )`. Similar to other recent changes, `K8sGateway.Bind( )` will proxy through to `Binder.Bind()`.

### How I've tested this PR:
:robot: tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/store/memory/store.go
```diff
@@ -222,7 +198,7 @@ func (s *Store) UpsertRoute(ctx context.Context, route store.Route, updateCondit
 
 	// bind to gateways
 	for _, gateway := range s.gateways {
-		gateway.TryBind(ctx, route)
+		gateway.Bind(ctx, route)
 	}
 
 	// sync the gateways to consul and route statuses to k8s
```
andrewstucki, line 204: Just for posterity sake, the optimization of only syncing based off of ingress gateway changes comes from trying to ensure the `Sync` call below this doesn't hammer Consul every time any resource is reconciled. In the last part of the refactor we do some smarter dirty checking to make sure we're only updating gateways whose state values have actually changed from their last stored state, so even if we temporarily un-gate the syncs an over-sync to Consul right now, it should be fixed soon enough.

@nathancoleman and I talked about that offline.

internal/k8s/reconciler/gateway.go
```diff
@@ -92,47 +92,80 @@ func (g *K8sGateway) Meta() map[string]string {
...
-		listeners = append(listeners, listener)
+// Bind returns the name of the listeners to which a route bound
+func (g *K8sGateway) Bind(ctx context.Context, route store.Route) []string {
+	k8sRoute, ok := route.(*K8sRoute)
+	if !ok {
+		return nil
 	}
 
-	return listeners
+	return newBinder(g.client, g.Gateway, g.GatewayState).Bind(ctx, k8sRoute)
```
nathancoleman, line 102: Here's where we consume the new `Binder` to bind the route to the gateway

internal/k8s/reconciler/route.go
```diff
@@ -197,32 +190,16 @@ func (r *K8sRoute) Validate(ctx context.Context) error {
 	return r.validator.Validate(ctx, r.RouteState, r.Route)
 }
 
-func (r *K8sRoute) OnBindFailed(err error, gateway store.Gateway) {
```
nathancoleman, line 200: These moved onto the `RouteState` at `internal/k8s/reconciler/state/route.go`

internal/store/memory/store.go
```diff
@@ -58,22 +57,6 @@ func (s *Store) GatewayExists(ctx context.Context, id core.GatewayID) (bool, err
 	return found, nil
 }
 
-func (s *Store) CanFetchSecrets(ctx context.Context, id core.GatewayID, secrets []string) (bool, error) {
```
nathancoleman, line 61: This moved onto the `store.Gateway` interface to limit the amount of logic the store has that isn't related to storing things

internal/store/memory/store.go
```diff
@@ -147,13 +130,12 @@ func (s *Store) sync(ctx context.Context, gateways ...*gatewayState) error {
...
+
 	syncGroup.Go(func() error {
 		return s.syncRouteStatuses(ctx)
 	})
-	if err := syncGroup.Wait().ErrorOrNil(); err != nil {
-		return err
-	}
-	return nil
+
+	return syncGroup.Wait().ErrorOrNil()
```
nathancoleman, line 138: Just an opportunity for simplification that I noticed

internal/store/memory/store.go
```diff
@@ -175,12 +157,6 @@ func (s *Store) SyncAtInterval(ctx context.Context) {
 			case <-ticker.C:
 				s.mutex.Lock()
 
-				// Force each gateway to sync its state even though listeners
-				// on the gateway may not be marked as needing a sync right now
-				for _, gateway := range s.gateways {
-					gateway.needsSync = true
-				}
-
```
nathancoleman, line 183: The sync is not gated the way it was before, so there's no need to "force the sync gate open" as this was doing before. The loop just needs to call `sync`. As Andrew mentions in a comment below in this file, we'll be adding some dirty checking logic in a followup PR to prevent an endless loop of reconciliation.

nathancoleman (commented):
Personal review

internal/envoy/handler.go
```diff
@@ -85,7 +86,12 @@ func (r *RequestHandler) OnStreamRequest(streamID int64, req *discovery.Discover
 	resources := req.GetResourceNames()
 
 	// check to make sure we're actually authorized to do this
-	allowed, err := r.registry.CanFetchSecrets(ctx, GatewayFromContext(ctx), resources)
+	gateway, err := r.store.GetGateway(ctx, GatewayFromContext(ctx))
```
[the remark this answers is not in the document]
andrewstucki: 💯

internal/k8s/reconciler/binder.go
```diff
@@ -0,0 +1,208 @@
...
+		Client:       client,
+		Gateway:      gateway,
+		GatewayState: state,
+	}
+}
+
+// Bind will attempt to bind the provided route to all listeners on the Gateway and
+// remove the route from any listeners that the route should no longer be bound to.
+// The latter is important for scenarios such as the route's parent changing.
+func (b *binder) Bind(ctx context.Context, route *K8sRoute) []string {
```
andrewstucki, line 44: I like this simplification 👍

internal/store/interfaces.go
```diff
@@ -32,10 +24,10 @@ type StatusTrackingGateway interface {
 // Gateway describes a gateway.
 type Gateway interface {
 	ID() core.GatewayID
-	Meta() map[string]string
-	ShouldUpdate(other Gateway) bool
-	Listeners() []Listener
-	ShouldBind(route Route) bool
+	Bind(ctx context.Context, route Route) []string
+	Remove(ctx context.Context, id string) error
+	Resolve() core.ResolvedGateway
```
andrewstucki, line 29: This is great that we're getting a lot closer

internal/store/interfaces.go
```diff
@@ -32,10 +24,10 @@ type StatusTrackingGateway interface {
...
+	CanFetchSecrets(ctx context.Context, secrets []string) (bool, error)
 }
 
 // ListenerConfig contains the common configuration
```
andrewstucki, line 33: I believe we should be able to drop all of the `Listener*` related stuff in this file at this point

internal/store/interfaces.go
```diff
@@ -77,22 +69,18 @@ type StatusTrackingRoute interface {
...
-	OnBound(gateway Gateway)
-	OnBindFailed(err error, gateway Gateway)
 	OnGatewayRemoved(gateway Gateway)
 }
 
 // Route should be implemented by all route
 // source integrations
 type Route interface {
 	ID() string
-	Resolve(listener Listener) core.ResolvedRoute
```
andrewstucki, line 89: 💯

internal/store/memory/gateway.go
```diff
@@ -1,153 +0,0 @@
-package memory
```
andrewstucki, line 1: 🎤 💧

andrewstucki (approved):
This is awesome that we're starting to get quite a bit closer, enough to actually drop some of the store implementation and old interfaces at this point. I can kind of see the end in sight 😅

internal/k8s/reconciler/listener.go
```diff
@@ -195,7 +195,7 @@ func (l *K8sListener) canBind(ctx context.Context, ref gwv1alpha2.ParentReferenc
 			return false, nil
 		}
 
-		if !route.MatchesHostname(l.listener.Hostname) {
```
andrewstucki, line 198: Last comment. I believe pretty much all of this file can get deleted too at this point since you're pulling/setting state from `Gateway` at this point, but you'll have to update the initializers if you drop this. Feel free to do so now or in a follow-up.

internal/store/interfaces.go
nathancoleman: 👍 totally! I'm saving that for my next PR just to keep this one down in size

internal/store/memory/gateway.go
nathancoleman: 🤣 🤣 🤣

internal/store/interfaces.go
nathancoleman: Closing the loop, `Listener*` code is removed in https://github.com/hashicorp/consul-api-gateway/pull/337

internal/k8s/reconciler/listener.go
nathancoleman: Handled in https://github.com/hashicorp/consul-api-gateway/pull/337

PR #336, opened 2022-09-01
(dark) Implement cert+key retrieval using Vault's KV API

nathancoleman:
#295 added dark code to support cert+key retrieval using Vault's PKI API.
This PR adds the same support but for the KV API.

### How I've tested this PR:
This code is not exposed via any API at the moment and so is only covered by unit tests.

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/vault/secrets.go
```diff
@@ -3,36 +3,97 @@ package vault
...
 const (
+	// PKI secret
 	queryKeyAltNames  = "altNames"
 	queryKeyIPSANs    = "ipSans"
 	queryKeyOtherSANs = "otherSans"
 	queryKeyTTL       = "ttl"
+
+	// Static secret
+	queryKeyCertKey       = "tlsCertKey"
+	queryKeyPrivateKeyKey = "tlsPrivateKeyKey"
```
nathancoleman, line 22: This reads a little weird but is parroting [the K8s Secret API](https://github.com/kubernetes/api/blob/d10499434b5204901e2681f883e988f35b63e2ba/core/v1/types.go#L6224-L6225). Not really glued to it though if someone has a different idea for naming

andrewstucki (approved):
nit suggestion just to change up our ridiculously overloaded usage of the word "key" -- aside from that looks almost identical to what I wrote independently, so going to 👍

internal/vault/secrets.go
nathancoleman: @andrewstucki any ideas here?

andrewstucki: yeah, sorry, issued my review but somehow the suggestion I made got dropped. Just so we don't talk about using a key to retrieve a key to get the proper keys within a key-value entry, maybe something like:

`queryPrivateKeyField = "tlsPrivateKeyField"` (i.e. use `field` instead of `key` in this context and just simplify `queryKey*` to `query*`)?

nathancoleman: I like that, will make that change! Thx

PR #338, opened 2022-09-02
Management API Stubs for Gateway endpoints

andrewstucki:

This adds stubs for some CRUD endpoints for the Gateway management APIs. It also makes some of the error messages for validation failures prettier.

Since there's nothing using this right now, it's not too delicate of a change. The additional stubs for routes and internal APIs can come in subsequent PRs.

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


.github/workflows/ci.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       run: |
         echo "Checking that code generation is up-to-date"
         go install github.com/golang/mock/mockgen@v1.6.0
-        go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.10.1
+        go install github.com/andrewstucki/oapi-codegen/cmd/oapi-codegen@indirection
```
andrewstucki, line 69: Just FYI, this is a temporary fork while this PR winds up getting worked out upstream: https://github.com/deepmap/oapi-codegen/pull/382 -- basically it makes it so we don't have an unneeded indirection for the `object`/`map` fields like `Meta` in the generated structs.

internal/api/schemas/v1.yaml
```diff
@@ -21,7 +21,86 @@ paths:
...
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+    post:
+      description: 'Creates a new gateway.'
+      operationId: createGateway
+      operationId: addGateway
```
nathancoleman, line 34:
```suggestion
      operationId: createGateway
```

internal/api/schemas/v1.yaml
```diff
@@ -21,7 +21,86 @@ paths:
...
+      requestBody:
+        description: Gateway to add
```
nathancoleman, line 36: be consistent w/ "add" vs. "create". I prefer "create" fwiw

internal/api/schemas/v1.yaml
```diff
@@ -21,7 +21,86 @@ paths:
...
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Gateway"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+  "/gateways/{namespace}/{name}":
```
nathancoleman, line 55: This endpoint structure implies that API consumers should always have a namespace for their gateway, even if that's something like `default`.

If this is the case, I'd prefer a more consistent and RESTful resource-based structure, if you're okay considering a "namespace" to be a resource/container:
- `GET /namespaces/{namespace}/gateways`
- `GET /namespaces/{namespace}/gateways/{name}`
-  `DELETE /namespaces/namespace/gateways/{name}`

Imagining we'd still need something like `GET /gateways` to retrieve all gateways that exist.

nathancoleman (commented):
Some initial thoughts. Great to see this happening! 🥳

internal/api/schemas/v1.yaml
nathancoleman: In VM world, where we don't have K8s namespaces, would you ever need to CRUD a namespace?

andrewstucki: It should always map back to a pre-existing Consul namespace, that said, I like the additional way namespaces are specified in your suggestion, it feels less nasty. Also, given that in OSS Consul you don't have namespaces (other than the magical `default`), I'm thinking we can do a mapping for

`/gateways/*` that is equivalent to `/namespaces/default/gateways/*`

if we want to we could add a query parameter to that endpoint to unscope the request from a given namespace (i.e. like an `namespaces=*` parameter or something of the sort.

WDYT?

internal/api/schemas/v1.yaml
andrewstucki: yeah, meant to use `createGateway` -- I'll clean up the dup

internal/api/schemas/v1.yaml
andrewstucki: Went ahead and did the above

jherschman (approved):
Looks great 🔥! Will set some time up next week to start the convo around how I can work with you on UX / API design for future roadmap items :)

internal/api/schemas/v1.yaml
nathancoleman: I like it!

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
     name: Mozilla Public License 2.0
     url: https://www.mozilla.org/en-US/MPL/2.0/
 paths:
+  "/namespaces/{namespace}/gateways":
```
nathancoleman, line 14: We'll need a `POST` on this path in order to create any gateways outside of the default namespace.

I think we'd just want `POST /gateways` to be an alias for `POST /namespaces/default/gateways`

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
...
+              schema:
+                "$ref": "#/components/schemas/GatewayPage"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+    post:
+      description: 'Creates a new gateway.'
```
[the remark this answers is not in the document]
andrewstucki: So, we have options here, we can either leave namespaces to _only_ be specified via the path, or (which I kind of prefer), have it be a part of the request body parameter (i.e. `Gateway.namespace`) since it'll need to be associated with the Gateway/Routes either way. I know it makes it so there's not complete symmetry between the namespaced/non-namespaced routes, but I figured in this case it'd probably be alright, at least for now.

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
...
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/GatewayPage"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
   "/gateways":
```
nathancoleman, line 38: I'm not sure I'd add this alias for the default namespace just yet - it feels like we're increasing our API surface area right off the bat when it may be unnecessary.

Honestly, I don't have any strong cons in my head other than the increased surface area, I just default to being very conservative with API design since we own it ~~forever~~ for a long time once it's out in the wild. Thoughts?

nathancoleman: Based on the query parameter below though, I see you're thinking of this as a way to query gateways across 2+ namespaces as well, so that may be a required feature that justifies the surface area.

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
     name: Mozilla Public License 2.0
     url: https://www.mozilla.org/en-US/MPL/2.0/
 paths:
+  "/namespaces/{namespace}/gateways":
+    get:
+      description: 'Returns all configured gateways in the given namespace.'
+      operationId: listNamespacedGateways
```
nathancoleman, line 17: "namespaced" describes every gateway IMO, so I'd go with `listGatewaysInNamespace` here

```suggestion
      operationId: listGatewaysInNamespace
```

internal/api/schemas/v1.yaml
andrewstucki: Yeah, so this would basically allow you to request either from default, or unscope from all namespaces -- since we _are_ going to have to support putting gateways/routes in multiple namespaces from the get-go, I'm fine with aliasing up-front.

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
...
+          description: gateway deleted
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+  "/gateways/{name}":
+    get:
+      description: Returns a gateway in the default namespace based off of its unique name.
```
nathancoleman, line 141: IIRC, mardown is valid here. I would do `"default"` just to indicate that we are talking about a specific, non-configurable namespace named "default"

internal/api/schemas/v1.yaml
```diff
@@ -11,17 +11,168 @@ info:
...
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Gateway"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+  "/namespaces/{namespace}/gateways/{name}":
```
nathancoleman, line 84: I could see us needing a `PUT` on this path, but I'm thinking you might be holding off on that until we discuss `PUT` vs. `PATCH` in the future?

internal/api/schemas/v1.yaml
nathancoleman: I believe we will need to list gateways in all namespaces. Are you thinking you reserve a keyword or something for that like `GET /gateways?namespaces=ALL`?

internal/api/schemas/v1.yaml
andrewstucki: Yeah, that and I'm also leaning towards making the `Create` endpoint more of an "upsert" endpoint in the long run (since you can identify gateways/routes without their path parameters due to them returning names/namespaces in their bodies) with PUT semantics more than anything -- which would eventually mean we could extend this to include PATCH, but wouldn't have to initially since we already have the ability to update.

internal/api/schemas/v1.yaml
andrewstucki: That's the thought, maybe making something like `*` the keyword just since no one is going to use that as the name of a namespace.

internal/api/schemas/v1.yaml
```diff
@@ -35,13 +186,99 @@ components:
...
+            "$ref": "#/components/schemas/Gateway"
+          minItems: 0
+        meta:
+          type: object
+    Gateway:
+      type: object
+      required:
+      - name
+      - namespace
+      - listeners
```
nathancoleman, line 206: Do we want to require `listeners` be defined in order to create a `Gateway`?

internal/api/schemas/v1.yaml
```diff
@@ -35,13 +186,99 @@ components:
...
+          type: array
+          items:
+            "$ref": "#/components/schemas/Listener"
+          minItems: 1
+        meta:
+          type: object
+    Listener:
+      type: object
+      required:
+      - hostname
```
nathancoleman, line 224: Do we want to require `hostname` be defined in order to create? IIRC, this is not required in K8s land

internal/api/schemas/v1.yaml
```diff
@@ -35,13 +186,99 @@ components:
...
+        privateKeyField:
+          type: string
+          minLength: 1
     Error:
       type: object
       required:
       - message
+      - code
       properties:
+        code:
```
nathancoleman, line 279: Do you expect this to align with standard HTTP response codes or some custom set of codes? Helpful to add a description here IMO

internal/api/schemas/v1.yaml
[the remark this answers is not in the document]
nathancoleman: I don't see a benefit of breaking from consistent patterns in this case. If every `Gateway` _must_ have a namespace, then I'd prefer we either
- always use the namespace in the path (and maybe alias just for the default namespace), or
- always use the namespace in the payload

internal/api/schemas/v1.yaml
nathancoleman: Fine punting on this for now

internal/api/schemas/v1.yaml
andrewstucki: I believe it _is_ required in K8s land, it just defaults to `*`, which I figure we should do as well.

internal/api/schemas/v1.yaml
andrewstucki: For now, yes, but I'm planning on modifying some of this stuff to be something more verbose and user-defined since you can already get this info via the HTTP response code, just haven't done it yet.

internal/api/schemas/v1.yaml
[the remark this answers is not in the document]
andrewstucki: So, that's the thing that's a bit wonky about how Consul works under the hood -- technically yes, namespaces _are_ always required, but in OSS if you set a namespace to anything other than the magic default namespace, you get an error. Furthermore, if you don't specify a namespace you wind up with the default one getting implicitly added for you.

So, in OSS, despite them being required you _never_ deal with namespaces. What this means from our perspective just so we don't introduce concepts that don't exist for a bunch of our users is that you should be able to create a Gateway/Route/whatever without specifying a namespace. If, however, you _do_ specify a namespace, it becomes part of its identifier.

Why I have a preference towards making this the only different endpoint is that it seems a bit odd to me to specify all other Gateway parameters in the request body and then merge the path-based namespace onto the payload.

internal/api/schemas/v1.yaml
```diff
@@ -229,6 +229,7 @@ components:
           type: string
         hostname:
           type: string
+          default: "*"
```
nathancoleman, line 232: 👍

nathancoleman (approved):
@andrewstucki and I discussed the way Consul namespaces work elsewhere and, understanding that we can't read a specific Gateway without including its namespace in the path, and that we also don't want to merge a namespace from the path and the request body on create operations, this API shape is what makes sense.

Over the coming weeks, we're going to add bits of this reasoning to the `description` fields in the API spec and spend time actually _using_ the API to see what does and doesn't feel good. This will have us in the best possible shape prior to releasing it out into the wild, at which point we own it 😜 .

PR #339, opened 2022-09-02
Add HTTPRoute stubs

andrewstucki:

THIS IS A STACKED PR SINCE IT REQUIRESR https://github.com/hashicorp/consul-api-gateway/pull/338 TO BE MERGED FIRST.

This re-orgs a bit of the handler structure and adds some stub routes for HTTPRoute CRUD operations.

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


jherschman (approved):
Looking goood 👍

internal/api/schemas/v1.yaml
```diff
@@ -271,6 +439,178 @@ components:
...
+      properties:
+        name:
+          type: string
+        namespace:
+          type: string
+        weight:
+          type: number
+        filters:
+          "$ref": "#/components/schemas/HTTPFilters"
+    HTTPFilters:
```
nathancoleman, line 517: Worth going ahead and adding path prefix replace filter since we support it in K8s land now?

nathancoleman (approved):

internal/api/schemas/v1.yaml
```diff
@@ -271,6 +439,178 @@ components:
...
+          minLength: 1
+        namespace:
+          type: string
+          minLength: 1
+    HTTPRoute:
+      type: object
+      required:
+      - name
+      - namespace
+      - gateways
```
nathancoleman, line 470: Analogous to `ParentRefs`, K8s doesn't require these be set. Thoughts?

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.CommonRouteSpec

internal/api/schemas/v1.yaml
andrewstucki: Good call, I'll do that in a follow-up.

internal/api/schemas/v1.yaml
andrewstucki: I'd prefer to keep this as required, at least initially. I've never quite understood why a route doesn't need a gateway in the K8s API -- think maybe due to the initial thoughts around things like route delegation?

Either way, it'd be a breaking change to all of a sudden disallow empty gateways, but it's a lot easier to make the validation _less_ strict.

mikemorris: Yea, spec flexibility will be useful here in the future for route delegation, GAMMA mesh binding, etc - requiring binding to a gateway for now makes sense to me.

PR #340, opened 2022-09-02
Add TCPRoute stubs

andrewstucki:

THIS IS A STACKED PR SINCE IT REQUIRES https://github.com/hashicorp/consul-api-gateway/pull/338 and https://github.com/hashicorp/consul-api-gateway/pull/339 TO BE MERGED FIRST.

This adds the route stubs for TCPRoutes that are corollary to HTTPRoutes

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


jherschman (approved):
🌟

internal/api/schemas/v1.yaml
```diff
@@ -611,6 +779,55 @@ components:
...
+          items:
+            "$ref": "#/components/schemas/TCPRoute"
+        meta:
+          type: object
+    TCPRoute:
+      type: object
+      required:
+      - name
+      - namespace
+      - gateways
```
nathancoleman, line 798: Analogous to `ParentRefs`, K8s doesn't require these be set. Thoughts?

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.CommonRouteSpec

nathancoleman (approved):
LGTM other than the one comment

internal/api/schemas/v1.yaml
andrewstucki: Same as the HTTPRoute stuff -- I'd prefer to keep this as required at least initially. We can always make the API more permissive if we so desire without it being a breaking change, but can't really go the other direction without a break -- but not entirely sure why the refs aren't required for K8s and I think(?) it has something to do with plans for later route delegation.

PR #342, opened 2022-09-02
(dark) Add storage "backend" interface, in-memory implementation

nathancoleman:
This is the first part to be broken out of #165 . This defines the new `Backend` interface that handles CRUD ops for Gateways and Routes. It's not yet consumed anywhere, so this code is dark. I will consume it in a future PR as part of the ongoing refactor.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/store/interfaces.go
+++ b/internal/store/interfaces.go
@@ -58,3 +58,27 @@ type Store interface {
 	UpsertRoute(ctx context.Context, route Route, updateConditionFn func(current Route) bool) error
 	Sync(ctx context.Context) error
 }
+
+// Backend is used for persisting and querying gateways and routes
+type Backend interface {
+	GetGateway(ctx context.Context, id core.GatewayID) ([]byte, error)
+	ListGateways(ctx context.Context) ([][]byte, error)
+	DeleteGateway(ctx context.Context, id core.GatewayID) error
+	UpsertGateways(ctx context.Context, gateways ...GatewayRecord) error
+	GetRoute(ctx context.Context, id string) ([]byte, error)
+	ListRoutes(ctx context.Context) ([][]byte, error)
+	DeleteRoute(ctx context.Context, id string) error
+	UpsertRoutes(ctx context.Context, routes ...RouteRecord) error
+}
+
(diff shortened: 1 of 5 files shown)


internal/store/interfaces.go
```diff
@@ -58,3 +58,27 @@ type Store interface {
 	UpsertRoute(ctx context.Context, route Route, updateConditionFn func(current Route) bool) error
 	Sync(ctx context.Context) error
 }
+
+// Backend is used for persisting and querying gateways and routes
+type Backend interface {
+	GetGateway(ctx context.Context, id core.GatewayID) ([]byte, error)
+	ListGateways(ctx context.Context) ([][]byte, error)
```
nathancoleman, line 65: @andrewstucki we're either going to have to make this list function more robust to support filtering to one or more namespaces, or we'll have to list all and filter further up in the application layer. Thoughts?

andrewstucki: Not entirely sure yet, for the first pass it's probably just easier to filter at a higher level, but I'm thinking we may want to move that logic into the actual store retrieval mechanism at some point to avoid the hit of the roundtrip/compute cost for Consul persistence.

nathancoleman: We'll leave as-is for now and revisit in the coming weeks

internal/store/interfaces.go
```diff
@@ -58,3 +58,27 @@ type Store interface {
...
+
+// GatewayRecord represents a serialized Gateway
+type GatewayRecord struct {
+	ID   core.GatewayID
+	Data []byte
+}
+
+// RouteRecord represents a serialized Route
+type RouteRecord struct {
+	ID   string
```
andrewstucki, line 82: Know this code is exactly what we currently leverage, but just calling out here that after the refactor is done we might want to switch this to be similar to `GatewayID` -- a compound struct with namespace/name info. Nothing to do for now though.

andrewstucki (approved):
LGTM

PR #343, opened 2022-09-03
Add internal endpoint stubs

andrewstucki:

This stubs the internal endpoint for bootstrapping a deployment along with an empty middleware to use for gateway token checking.

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/api/internal/server.go
```diff
@@ -0,0 +1,57 @@
...
+var _ ServerInterface = &Server{}
+
+type Server struct {
+	logger       hclog.Logger
+	consulClient *api.Client
+}
+
+// TODO(andrew): most of this is boilerplate that should be generated
+
+func NewServer(url string, consulClient *api.Client, logger hclog.Logger) http.Handler {
```
sarahalsmiller, line 25: There are three different functions named NewServer across these packages that seem to do slightly different variations of server/handler setup, and I'd appreciate some header comments clarifying the difference between the three, but otherwise seems solid to me.

sarahalsmiller (approved):

internal/api/internal/server.go
andrewstucki: Yep, we can add some docstrings as we go. The main thing is that we're composing two APIs together. Each API, internal & v1 are structured as their own API "servers" in that you can easily run them as standalone services (as we do in the integration-level tests here). The top-level `api/server.go` implementation just composes them together and should pretty much be what we use as the main thing we invoke in our entrypoint.

andrewstucki: The clients, btw are pretty much structured the same way. Mind if I just follow-up with some better doc comments as I flesh out the scaffolding?

PR #344, opened 2022-09-03
Add stubs for gateway client

andrewstucki:

So this is a huge diff because it's generating a few thousand lines of client code. In addition, the templates generate a slightly more ergonomic wrapper around the generated client that has the following methods:

```go
// Create methods
func (c *APIClient) CreateGateway(ctx context.Context, body Gateway) (*Gateway, error)
func (c *APIClient) CreateHTTPRoute(ctx context.Context, body HTTPRoute) (*HTTPRoute, error)
func (c *APIClient) CreateTCPRoute(ctx context.Context, body TCPRoute) (*TCPRoute, error)

// List methods
func (c *APIClient) ListGatewaysInNamespace(ctx context.Context, namespace string) (*GatewayPage, error)
func (c *APIClient) ListHTTPRoutesInNamespace(ctx context.Context, namespace string) (*HTTPRoutePage, error)
func (c *APIClient) ListTCPRoutesInNamespace(ctx context.Context, namespace string) (*TCPRoutePage, error)

// List methods with query parameters
func (c *APIClient) ListGateways(ctx context.Context, params ...ListGatewaysParams) (*GatewayPage, error)
func (c *APIClient) ListHTTPRoutes(ctx context.Context, params ...ListHTTPRoutesParams) (*HTTPRoutePage, error)
func (c *APIClient) ListTCPRoutes(ctx context.Context, params ...ListTCPRoutesParams) (*TCPRoutePage, error)

// Get methods
func (c *APIClient) GetGateway(ctx context.Context, name string) (*Gateway, error)
func (c *APIClient) GetHTTPRoute(ctx context.Context, name string) (*HTTPRoute, error)
func (c *APIClient) GetTCPRoute(ctx context.Context, name string) (*TCPRoute, error)
func (c *APIClient) GetGatewayInNamespace(ctx context.Context, namespace string, name string) (*Gateway, error)
func (c *APIClient) GetHTTPRouteInNamespace(ctx context.Context, namespace string, name string) (*HTTPRoute, error)
func (c *APIClient) GetTCPRouteInNamespace(ctx context.Context, namespace string, name string) (*TCPRoute, error)

// Delete methods (notice the no return value since we just do 204s with no content on delete)
func (c *APIClient) DeleteGateway(ctx context.Context, name string) error
func (c *APIClient) DeleteHTTPRoute(ctx context.Context, name string) error
func (c *APIClient) DeleteTCPRoute(ctx context.Context, name string) error
func (c *APIClient) DeleteGatewayInNamespace(ctx context.Context, namespace string, name string) error
func (c *APIClient) DeleteHTTPRouteInNamespace(ctx context.Context, namespace string, name string) error
func (c *APIClient) DeleteTCPRouteInNamespace(ctx context.Context, namespace string, name string) error
```

These methods do some additional error handling since the underlying openapi client doesn't consider anything outside of the 500 range an "error", so if we have an `Error` structure that gets unmarshaled by the underlying client (say due to an authentication issue), using the wrappers, we now treat that as an error that can be differentiated via using something like `errors.As/Is` on the returned `error`.

The only slightly funny thing here, that I've seen plenty of other times is the abuse of variadics to allow for optional specification of query parameters, in this case only the first set of parameters passed in the variadic is used.

Most of the templates aren't really necessary to understand for the sake of review, but it is probably worth taking a look at the code that gets generated in `internal/api/v1/zz_generated_extensions.go`. I also added some notes for some places where I think we can generate some additional code later just to avoid mixing what should be pretty standard boilerplate and code that we write.

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/api/templates/client.tmpl
```diff
@@ -0,0 +1,29 @@
+// Start of User Generated Client
```
sarahalsmiller, line 1: Looks good, quick question, is this template needed or does the template with responses make it redundant?

sarahalsmiller (approved):

internal/api/templates/client.tmpl
andrewstucki: So, unfortunately this is a bit of a hack for now, what's going on is that the `client` target for `oapi-codegen` **_requires both template files_**, so basically I just separated the more "static" stuff that is boilerplate into the `client.tmpl` file and the more dynamic stuff that actually reflects on the schema in `client-with-responses.tmpl`.

internal/api/v1/server.go
```diff
@@ -12,6 +11,7 @@ import (
 )
 
 //go:generate oapi-codegen -config ../schemas/v1.config.yaml ../schemas/v1.yaml
+//go:generate oapi-codegen -old-config-style -package v1 -generate client -templates ../templates -o zz_generated_extensions.go ../schemas/v1.yaml
```
andrewstucki, line 14: Just to be clear what's going on here -- we're doing a second pass on the "client" generation target, but this time with overriding the templates used with our custom templates to generate the wrappers. `-old-config-style` is used since the "new style" strangely requires defining the templates inline in the yaml file, which I'd rather not do.

So, all of the normal stuff, including client and server impls gets generated by the first `go:generate` target, the second re-runs the codegen with custom templates to add our desired extensions for client ergonomics.

PR #350, opened 2022-09-08
(dark) Add Marshaler for serializing + deserializing routes + gateways

nathancoleman:
### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/route.go
+++ b/internal/k8s/reconciler/route.go
@@ -1,12 +1,14 @@
 package reconciler
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 
 	"github.com/hashicorp/go-hclog"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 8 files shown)


internal/k8s/reconciler/route.go
```diff
@@ -177,15 +186,15 @@ func (r *K8sRoute) Parents() []gwv1alpha2.ParentReference {
 	return nil
 }
 
-func (r *K8sRoute) Validate(ctx context.Context) error {
+func (r *K8sRoute) validate(ctx context.Context) error {
```
andrewstucki, line 189: Did we want to strip this out at this point? I'm all for keeping the method surface area for these fairly slim just so we don't wind up fattening up the route and gateway wrappers in the future..

andrewstucki (approved):
Only one real question from me

internal/k8s/reconciler/route.go
nathancoleman: It's on my list, but I'm planning to do it in a separate PR as similar changes will need to happen for the gateway

PR #351, opened 2022-09-08
(dark) Include marshaling funcs when generating statuses

nathancoleman:
This is the second part to be broken out of https://github.com/hashicorp/consul-api-gateway/pull/165 . This adds marshal/unmarshal funcs for the generated statuses by extending the generator.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+**/*zz_generated_*.go linguist-generated=true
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+**/*zz_generated_*.go linguist-generated=true
```
nathancoleman, line 1: > **Note** This will indicate to GitHub ([docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)) that `zz_generated_*.go` files are generated and so they aren't loaded by default.

andrewstucki (approved):

PR #356, opened 2022-09-09
VM controller health

andrewstucki:

This adds part of the implementation for returning controller/gateway status, the gateway part of this is going to require introspecting our gateways to figure out the list of services that we need to query from Consul for their health, which is going to require the persistence layer be in place. Once that's there I'd like to add some robust tests for this -- right now I'm still considering this the scaffolding phase.

### How I've tested this PR:

Manually.

Here is the output of the command when running a controller in a different terminal:

```
➜  consul-api-gateway git:(vm-controller-health) ./consul-api-gateway health
Successfully retrieved controller health: &{[{c656a295-5455-4c5e-b81e-82557062a8a4 api-gateway-controller passing}] []}
```

with matching Consul status:
Screen Shot 2022-09-09 at 10 47 48 AM

And when running two controllers:

```
➜  consul-api-gateway git:(vm-controller-health) ./consul-api-gateway health
Successfully retrieved controller health: &{[{97f30700-bb63-43a3-bd78-8859eeef4285 api-gateway-controller passing} {c656a295-5455-4c5e-b81e-82557062a8a4 api-gateway-controller passing}] []}
```

with matching Consul status:
Screen Shot 2022-09-09 at 10 47 33 AM

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/api/schemas/v1.yaml
```diff
@@ -828,6 +845,46 @@ components:
...
+      - name
+      - id
+      - status
+      properties:
+        name:
+          type: string
+        id:
+          type: string
+        status:
+          type: string
```
nathancoleman, line 887: Makes sense to split this for future-proofing. Just curious if you have any ideas for how they might diverge at the moment?

internal/commands/health/command.go
```diff
@@ -0,0 +1,55 @@
...
+	health, err := client.V1().Health(c.Context())
+	if err != nil {
+		return c.Error("sending the request", err)
+	}
+
+	return c.Success(fmt.Sprintf("Successfully retrieved controller health: %v", health))
+}
+
+const (
+	synopsis = "Gets the health of all registered Consul Gateway API controllers and deployments"
```
nathancoleman, line 47: Looks like we're using "gateway" in the health API

```suggestion
	synopsis = "Gets the health of all registered Consul Gateway API controllers and gateways"
```

nathancoleman (approved):

internal/commands/health/command.go
andrewstucki: 👍 I'll also fix my tendency to swap `Consul API Gateway` to `Consul Gateway API` 😞

internal/api/schemas/v1.yaml
andrewstucki: not entirely sure -- I/whoever else might finish up the implementation (with the gateway querying part of this) should make sure to try and review the reqs to see if there's any information in general that we want to convey via our CLI tool. We should also probably check in with @jherschman once we have some of that baseline info displayed just to make sure everything at the command line is both useful and pretty.

internal/api/apiinternal/bootstrap.go
```diff
@@ -3,7 +3,7 @@ package apiinternal
 import "net/http"
 
 func (s *Server) Bootstrap(w http.ResponseWriter, r *http.Request) {
-	// do the actual bootstrap retrieval here
+	// do the actual bootstrap retrieval here along with token header enforcement
```
mikemorris, line 6: Nit, but it's helpful to prefix these with `TODO:` for finding later, some editors will even pick up that syntax for making a TODO list.

PR #357, opened 2022-09-09
Move gateway + route validation up into reconcile manager

nathancoleman:
This is the fourth part to be broken out of https://github.com/hashicorp/consul-api-gateway/pull/165 . This removes the validation entrypoint from the route + gateway themselves to further simplify them and instead calls the validator directly from the `ReconcileManager` before doing the upsert into the store.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/manager.go
+++ b/internal/k8s/reconciler/manager.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/consul-api-gateway/internal/common"
 	"github.com/hashicorp/consul-api-gateway/internal/core"
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/gatewayclient"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/reconciler/validator"
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/service"
 	"github.com/hashicorp/consul-api-gateway/internal/k8s/utils"
 	"github.com/hashicorp/consul-api-gateway/internal/store"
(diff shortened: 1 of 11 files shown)


internal/k8s/reconciler/manager.go
```diff
@@ -219,18 +225,18 @@ func (m *GatewayReconcileManager) upsertRoute(ctx context.Context, r Route, id s
...
+	} else if !managed {
 		return nil
 	}
 
 	// Calling validate outside of the upsert process allows us to re-resolve any
 	// external references and set the statuses accordingly. Since we actually
 	// have other object updates triggering reconciliation loops, this is necessary
 	// prior to dirty-checking on upsert.
-	if err := route.validate(ctx); err != nil {
+	if err = m.routeValidator.Validate(ctx, route.RouteState, route.Route); err != nil {
```
andrewstucki, line 236: Is the plan to move the state initialization into the `Validate` call like (I believe) the previous refactor form had? Just wondering since the separation of the two main fields on a route/gateway seems a bit odd here.

Something more akin to:

```
route := &v1beta1.HTTPRoute{}
state, err := routeValidator.Validate(ctx, route)
...
store.Upsert(ctx, NewK8sRoute(route, state))
...
```

seems like it might be a bit of a cleaner abstraction -- the initial state coming from the validation of the original Kubernetes object and then the composing of the two into our "stateful route" right before upsert.

andrewstucki (approved):
Seems fine to me in that it's incremental progress, but 1 question for clarification/suggestion.

internal/k8s/reconciler/manager.go
nathancoleman: It felt weird to me to have `.Validate()` spit out the state for a route it but thinking on it more, I believe it might just be the function name that bothers me. `Validate()` feels like it should answer "is this thing valid? true or false", not build a state object for the route. I'll spend some time this afternoon pondering, and maybe it will click or I can think up a name that resonates more.

andrewstucki: Yeah, I'm fine with a change in the name -- the main thing in my mind though is that the majority (with some exceptions since it normalizes the remote refs, though resolving those are a side effect of making sure the refs are actually valid) of what the validators do _is_ handle validation. The problem is with the fact that in the Kubernetes world giving a response to say, "this thing is invalid" is inherently a stateful operation because it's tied to the Status field in the object that needs to be updated.

So, don't mind if you change the name, but at least in the Kubernetes world, my mind doesn't separate out validation from what otherwise are stateful operations.

PR #358, opened 2022-09-09
Refactor binder interface to be reusable for multiple gateways

nathancoleman:
This is the fifth part to be broken out of https://github.com/hashicorp/consul-api-gateway/pull/165 . This restructures the `Binder` interface so that a single binder is reusable for multiple gateways instead of wrapping a single gateway.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/k8s/reconciler/binder.go
+++ b/internal/k8s/reconciler/binder.go
@@ -25,29 +25,23 @@ const (
 // binder wraps a Gateway and the corresponding GatewayState and encapsulates
 // the logic for binding new routes to that Gateway.
 type binder struct {
-	Client       gatewayclient.Client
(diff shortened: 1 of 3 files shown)


andrewstucki (approved):

internal/k8s/reconciler/binder.go
```diff
@@ -25,40 +25,34 @@ const (
 // binder wraps a Gateway and the corresponding GatewayState and encapsulates
 // the logic for binding new routes to that Gateway.
```
mikemorris, line 26: Should this comment be updated?

PR #359, opened 2022-09-09
VM initial runtime validation

andrewstucki:

This plumbs in a `validator` on Gateway/Route put operations.

For routes all we do is check to make sure that their services that they reference exist.

For gateways we do some checks about:

-  listener TLS configuration that corresponds 1-1 with the logic we currently have in K8s (though the cipher check only when min version is set seems odd to me?)
- TLS certificate references, making sure they not only exist and only have a single field set, but that, in the case of something like Vault KV, the secret + proper fields exist
- Listeners have unique ports and names

As we add the actual binding and storage, I plan on leveraging more e2e style tests, but this mocks out much of the more complex validation logic in the API unit tests (while having coverage in the validator unit tests) for the time being.

### How I've tested this PR:

Unit tests + manual testing

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/vm/validation.go
```diff
@@ -0,0 +1,225 @@
...
+		errs.Errors = append(errs.Errors, v.validateListerTLS(i, listener.Tls)...)
+
+		// validate TLS certs
+		errs.Errors = append(errs.Errors, v.validateListerTLSCertificates(ctx, i, listener.Tls)...)
+	}
+
+	return errs.ErrorOrNil()
+}
+
+func (v *Validator) validateListerTLS(index int, tls *v1.TLSConfiguration) []error {
```
sarahalsmiller, line 68: Should this be validateListenerTLS? Same question for validateListerTLSCertificates? Logic seems fine though so I'm not gonna block.

sarahalsmiller (approved):

internal/vm/validation.go
andrewstucki: 😊 ... spelling 😞

internal/vm/validation.go
```diff
@@ -0,0 +1,225 @@
...
+func NewValidator(logger hclog.Logger, vaultClient vault.KVClient, client *consulapi.Client) *Validator {
+	return &Validator{
+		client:      client,
+		vaultClient: vaultClient,
+		logger:      logger,
+	}
+}
+
+// ValidateGateway validates a Gateway definition
+func (v *Validator) ValidateGateway(ctx context.Context, gateway *v1.Gateway) error {
```
mikemorris, line 33: It feels like some of this logic is duplicated from the k8s validation and it would be desirable to ensure they stay in sync - should substantial parts of this (and the corresponding k8s logic) be moved out to a common layer above the `vm` and `k8s` packages?

PR #369, opened 2022-09-14
Replace old store w/ new store

nathancoleman:
> **Note** Please review https://github.com/hashicorp/consul-api-gateway/pull/383 first as this PR is based on top of that changeset

This is the last substantial PR of the refactor in #165, only a final cleanup pass remains.

The new `Store` has been built alongside the old `Store` in previous PRs. This PR replaces all uses of the old `Store` with the new and then deletes the old implementation.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 


internal/commands/controller/command.go
```diff
@@ -161,24 +162,20 @@ func (c *Command) Run(args []string) (ret int) {
...
-	group.Go(func() error {
-		store.SyncAtInterval(groupCtx)
-		return nil
+	store := store.New(store.Config{
+		Adapter:       consulAdapters.NewSyncAdapter(logger.Named("consul-adapter"), client),
+		Backend:       store.NewMemoryBackend(),
+		Binder:        nil, // TODO VM runtime implementation
+		Logger:        logger.Named("store"),
+		Marshaler:     nil, // TODO VM runtime implementation
+		StatusUpdater: nil, // TODO VM runtime implementation
```
nathancoleman, line 171: These TODOs will not be part of the refactor as this command is new and purely for VM support

internal/commands/server/server.go
```diff
@@ -61,12 +61,10 @@ func RunServer(config ServerConfig) int {
 	}
 
 	adapter := consulAdapters.NewSyncAdapter(config.Logger.Named("consul-adapter"), consulClient)
-	store := memory.NewStore(memory.StoreConfig{
-		Adapter: adapter,
-		Logger:  config.Logger.Named("state"),
-	})
+	store := store.New(k8s.StoreConfig(adapter, controller.Client(), config.Logger, *config.K8sConfig))
```
nathancoleman, line 64: This is the core change of this PR: replace the old store with the new

internal/envoy/sds_test.go
```diff
@@ -259,12 +262,13 @@ func runTestServer(t *testing.T, ca []byte, registryFn func(*gomock.Controller)
 		Name: "test",
 	}, time.Now(), nil)
 
-	sds := NewSDSServer(hclog.NewNullLogger(), fetcher, secretClient, memory.NewStore(memory.StoreConfig{}))
-	sds.bindAddress = serverAddress
-	sds.protocol = "unix"
+	var store store.Store = storeMocks.NewMockStore(ctrl)
 	if registryFn != nil {
-		sds.store = registryFn(ctrl)
+		store = registryFn(ctrl)
```
nathancoleman, line 267: Some tests will provide a func to build the store themselves. If they do, use it instead of the mock store

internal/k8s/config.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/hashicorp/go-hclog"
+
+	"github.com/hashicorp/consul-api-gateway/internal/core"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/gatewayclient"
+	"github.com/hashicorp/consul-api-gateway/internal/k8s/reconciler"
+	"github.com/hashicorp/consul-api-gateway/internal/store"
+)
+
+func StoreConfig(adapter core.SyncAdapter, client gatewayclient.Client, logger hclog.Logger, config Config) store.Config {
```
nathancoleman, line 12: This just exists to simplify the construction of the dependencies (deployer, status updater, etc.)

internal/k8s/controller.go
```diff
@@ -141,14 +141,18 @@ func (k *Kubernetes) SetStore(store store.Store) {
 	k.store = store
 }
 
+func (k *Kubernetes) Client() gatewayclient.Client {
```
nathancoleman, line 144: We need external access to the `gatewayclient` after constructing the controller in `internal/commands/server/server.go`; otherwise, we have to duplicate a bunch of kube client init logic in both places.

internal/k8s/reconciler/binder.go
```diff
@@ -15,15 +15,17 @@ import (
...
 const (
 	// NamespaceNameLabel represents that label added automatically to namespaces is newer Kubernetes clusters
 	NamespaceNameLabel = "kubernetes.io/metadata.name"
 )
 
-// binder wraps a Gateway and the corresponding GatewayState and encapsulates
-// the logic for binding new routes to that Gateway.
+var _ store.Binder = (*binder)(nil)
+
+// binder wraps encapsulates the logic for binding new Route(s) to a Gateway.
```
nathancoleman, line 28:
```suggestion
// binder encapsulates the logic for binding new Route(s) to a Gateway.
```

internal/k8s/reconciler/binder.go
```diff
@@ -35,34 +37,65 @@ func NewBinder(client gatewayclient.Client) *binder {
 // Bind will attempt to bind the provided route to all listeners on the Gateway and
 // remove the route from any listeners that the route should no longer be bound to.
 // The latter is important for scenarios such as the route's parent changing.
-func (b *binder) Bind(ctx context.Context, gateway *K8sGateway, route *K8sRoute) []string {
+func (b *binder) Bind(ctx context.Context, gateway store.Gateway, route store.Route) bool {
```
nathancoleman, line 40: The signature for this function changes a bit to return whether the gateway or route was modified instead of the list of bound listener names

internal/k8s/reconciler/binder.go
```diff
@@ -35,34 +37,62 @@ func NewBinder(client gatewayclient.Client) *binder {
...
+}
+
+func (b *binder) Unbind(_ context.Context, gateway store.Gateway, route store.Route) bool {
+	k8sGateway, k8sRoute := gateway.(*K8sGateway), route.(*K8sRoute)
+
+	removed := false
+	for _, listenerState := range k8sGateway.GatewayState.Listeners {
+		if _, ok := listenerState.Routes[k8sRoute.ID()]; ok {
+			removed = true
+			b.removeGatewayStatus(k8sRoute, k8sGateway)
```
nathancoleman, line 79: `Unbind()` is new but is important because it removes the route from the gateway's status.

internal/k8s/reconciler/factory.go
```diff
@@ -10,6 +10,7 @@ import (
 	apigwv1alpha1 "github.com/hashicorp/consul-api-gateway/pkg/apis/v1alpha1"
 )
 
+// TODO Remove Factory as it's no longer necessarygithub
```
nathancoleman, line 13: `Factory` is no longer really necessary and will be removed in the final cleanup pass

internal/k8s/reconciler/gateway.go
```diff
@@ -74,27 +39,6 @@ func (g *K8sGateway) ID() core.GatewayID {
 	}
 }
 
-// Bind returns the name of the listeners to which a route bound
-// TODO Remove
-func (g *K8sGateway) Bind(ctx context.Context, route store.Route) []string {
```
nathancoleman, line 79: This logic and the related logic below is now handled elsewhere and no longer needed since we're removing the `StatusTrackingGateway` interface

internal/k8s/reconciler/gateway_test.go
```diff
@@ -43,109 +37,3 @@ func TestGatewayID(t *testing.T) {
 	})
 	require.Equal(t, internalCore.GatewayID{Service: "name", ConsulNamespace: "consul"}, gateway.ID())
 }
-
-func TestGatewayTrackSync(t *testing.T) {
```
nathancoleman, line 47: Removing test code related to removed functions above in `gateway.go`

internal/k8s/reconciler/manager.go
```diff
@@ -91,27 +90,26 @@ func NewReconcileManager(config ManagerConfig) *GatewayReconcileManager {
...
 			Resolver:       resolver,
 		}),
+		logger:           config.Logger,
+		gatewayClasses:   NewK8sGatewayClasses(config.Logger.Named("gatewayclasses"), config.Client),
+		gatewayValidator: validator.NewGatewayValidator(config.Client),
+		namespaceMap:     make(map[types.NamespacedName]string),
+		routeValidator:   validator.NewRouteValidator(resolver, config.Client),
+		sdsHost:          config.SDSHost,
+		sdsPort:          config.SDSPort,
+		store:            config.Store,
```
nathancoleman, line 112: These got difficult to track across code areas and PRs, so I sorted them alphabetically

internal/store/interfaces.go
```diff
@@ -10,78 +10,43 @@ import (
...
 // Route should be implemented by all route
 // source integrations
 type Route interface {
 	ID() string
 }
 
-// Store is used for persisting and querying gateways and routes
-// TODO Remove
-type Store interface {
+type ReadStore interface {
```
nathancoleman, line 26: I separated into composable `ReadStore` and `WriteStore` interfaces. This helped me reason with where we're sending the store in our program entrypoint (envoy middleware, for example, doesn't need to update the store at all - only read it)

internal/store/interfaces.go
```diff
@@ -10,78 +10,43 @@ import (
...
 	DeleteRoute(ctx context.Context, id string) error
 
 	SyncAllAtInterval(ctx context.Context)
 }
 
+// Store is used for persisting and querying gateways and routes
+type Store interface {
+	ReadStore
+	WriteStore
+}
```
nathancoleman, line 48: Still composable when we need both

internal/store/store.go
```diff
@@ -73,7 +71,7 @@ func (s *lockingStore) UpsertGateway(ctx context.Context, gateway Gateway, updat
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
-	current, err := s.GetGateway(ctx, gateway.ID())
+	current, err := s.getGateway(ctx, gateway.ID())
```
nathancoleman, line 74: Use the non-locking version to avoid an infinite lock contention

internal/store/store.go
```diff
@@ -89,10 +87,7 @@ func (s *lockingStore) UpsertGateway(ctx context.Context, gateway Gateway, updat
 	}
 
 	// Attempt to bind all Route(s) to the Gateway in case any are newly-able to bind
-	_, boundRoutes, err := s.bindAll(ctx, []Gateway{gateway}, routes)
-	if err != nil {
-		return err
-	}
+	_, boundRoutes := s.bindAll(ctx, []Gateway{gateway}, routes)
```
nathancoleman, line 90: `bindAll` and the binding functions that make it up never actually returned error values. Since this is an internal function, we can always add back if we need to later, but this made things easier to reason about for now

internal/testing/e2e/consul.go
```diff
@@ -423,8 +423,8 @@ func ConsulClient(ctx context.Context) *api.Client {
 	return mustGetTestEnvironment(ctx).consulClient
 }
 
-func ConsulCA(ctx context.Context) []byte {
-	return mustGetTestEnvironment(ctx).ca
+func ConsulCA(ctx context.Context) string {
```
nathancoleman, line 426: Everything that used this func was casting to `string`, so I just did it at the source

nathancoleman (commented):
Personal review

sarahalsmiller (approved):
Looks solid to me, appreciate your annotations a lot.

PR #371, opened 2022-09-15
Register via TTL check for nodes with no agents

andrewstucki:

This swaps the health check that's registered as part of the exec process to use TTL-based checks in Consul.

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


nathancoleman:
> Andrew Stucki
  [13 hours ago](https://hashicorp.slack.com/archives/C02V3V13SB1/p1666056921531719?thread_ts=1666038506.121499&cid=C02V3V13SB1)
Sorry, just saw this -- actually chatted with Mike about that offline awhile ago, there's no downside

PR #376, opened 2022-09-16
remove acl tokens on shutdown in exec

kschoche:
We currently provision ACL tokens for each gateway on startup but do not clean them up this leads to abandoned ACL tokens for each gateway that restarts, every time the pod restarts.

### How I've tested this PR:
Manually deployed consul with ACLs enabled and api-gateway based on this branch, deployed a sample api-gw instance and checked the ACL tokens in Consul in the UI while deleting the gateway pod repeatedly.
Added a unit test.

- I can add a happy-path unit test but didnt see that as a pattern yet, happy to do this.
- Unit + manual test
- 👀 

### Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.changelog/376.txt
+++ b/.changelog/376.txt
@@ -1,0 +1,3 @@
+```release-note:bug-fix
+Delete gateway ACL tokens on shutdown so they are not orphaned after being provisioned at startup.
+```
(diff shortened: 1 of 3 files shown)


nathancoleman:
@kschoche I moved up to hashicorp/consul-api-gateway since our enterprise tests can't run on a fork.

See #377

PR #383, opened 2022-09-20
Clean up before replacing old store w/ new

nathancoleman:
Some general cleanup and small additions from the series of refactor PRs that I've broken #165 into. This is to help make the next PR where I replace our current store with the new one as small and focused as possible.

### How I've tested this PR:
🤖 tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/core/http.go
```diff
@@ -129,6 +131,10 @@ func (r HTTPRoute) GetType() ResolvedRouteType {
 	return ResolvedHTTPRouteType
 }
 
+func (r HTTPRoute) Marshal() ([]byte, error) {
```
nathancoleman, line 134: You'll see several marshal/unmarshal additions in this PR that weren't needed up until the actual store replacement which I have completed in a separate branch

internal/core/resolved.go
```diff
@@ -17,6 +22,50 @@ type ResolvedRoute interface {
...
 	GetNamespace() string
+	Marshal() ([]byte, error)
+}
+
+type marshaledRoute struct {
+	Type  ResolvedRouteType
+	Route json.RawMessage
+}
+
+// TODO Is this the best location for this func?
```
nathancoleman, line 33: I feel like we have marshaling structs + funcs spread across the codebase a little too much, but I've left this as something for the final round of cleanup after we've transitioned to the new `Store`

internal/k8s/reconciler/gatewayclass.go
```diff
@@ -126,7 +126,7 @@ func (c *K8sGatewayClass) Validate(ctx context.Context) error {
 func (c *K8sGatewayClass) SyncStatus(ctx context.Context) error {
 	current := c.class.Status.Conditions
 	conditions := c.status.Conditions(c.class.Generation)
-	if !conditionsEqual(conditions, current) {
+	if !rstatus.ConditionsEqual(conditions, current) {
```
nathancoleman, line 129: This replaces the final usage of some old equality testing funcs that you'll now see removed below

internal/k8s/reconciler/utils.go
```diff
@@ -45,89 +42,3 @@ func hostnamesMatch(a gwv1alpha2.Hostname, b gwv1beta1.Hostname) bool {
 
 	return string(a) == string(b)
 }
-
-func asJSON(item interface{}) string {
```
nathancoleman, line 49: These are the now-unused funcs that I referenced removing above

internal/k8s/reconciler/utils_test.go
```diff
@@ -36,134 +35,3 @@ func TestHostnamesMatch(t *testing.T) {
 	require.False(t, hostnamesMatch("*.b.test", "a.c.test"))
 	require.True(t, hostnamesMatch("a.b.test", "a.b.test"))
 }
-
-func TestConditionEqual(t *testing.T) {
```
nathancoleman, line 40: Tests for now-unused funcs

internal/k8s/reconciler/validator/route_test.go
```diff
@@ -52,7 +51,8 @@ func TestRouteValidate(t *testing.T) {
 		},
 	})
 	require.NoError(t, err)
-	require.True(t, routeState.ResolutionErrors.Empty())
+	require.NotNil(t, routeState)
+	assert.True(t, routeState.ResolutionErrors.Empty())
```
nathancoleman, line 55: Changes in this file are just improving our test hygiene based on a linter run that I did locally

internal/k8s/service/resolver.go
```diff
@@ -165,22 +179,16 @@ type ResolvedReference struct {
 	Type      ResolvedReferenceType
 	Reference *BackendReference
 	Consul    *ConsulService
-	object    client.Object
```
nathancoleman, line 168: `object` was only here for the `Item()` func below which is now completely unused

nathancoleman (commented):
Personal review

internal/store/store.go
```diff
@@ -230,6 +234,32 @@ func (s *lockingStore) DeleteRoute(ctx context.Context, id string) error {
 	return s.syncGateways(ctx, modifiedGateways)
 }
 
+// SyncAllAtInterval syncs all known Gateway(s) and Route(s) into Consul
+// at a regular interval. This function is blocking.
+func (s *lockingStore) SyncAllAtInterval(ctx context.Context) {
```
nathancoleman, line 239: This implements the background sync of all resources from the store added in #278 to deal with resources deleted out of band (such as a k8s Service, etc.)

mikemorris: Assuming this will be covered by tests when it's actually consumed?

internal/core/resolved.go
mikemorris: It feels like there _should_ be a cleaner way to implement this (don't the plain `Marshal` and `Unmarshal` function names have some magic behavior? Is that maybe only implemented by [protobuf](https://github.com/golang/protobuf/issues/955)?) but I don't have any concrete suggestions on _how_ though, so deferring any recommendation here for now.

internal/core/http.go
```diff
@@ -129,6 +131,10 @@ func (r HTTPRoute) GetType() ResolvedRouteType {
...
+	return json.Marshal(r)
```
sarahalsmiller, line 135: Is there a reason we need these instead of just calling json.Marshal(r) directly? Just convenience?

mikemorris (approved):

sarahalsmiller (approved):
Looks solid to me I just had the one question.

internal/store/store.go
nathancoleman: This is already covered by an e2e test added in #278

PR #384, opened 2022-09-20
api: add schema and stubs for bootstrap token CRUD

mikemorris:

- Add initial schema and stubs for bootstrap token CRUD operations.
- Add a new `make gen` command to install oapi-codegen if necessary and regenerate generated code and golden test files.

### How I've tested this PR:

Running `make gen` and `go test ./...` in `internal/api/v1` to make sure I didn't accidentally break anything.

Is there a less verbose way to nest sub-paths like the token CRUD operations under `/gateway/{name}/` in the OpenAPI schema?

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


.github/workflows/ci.yml
```diff
@@ -65,11 +65,7 @@ jobs:
     - name: Clean Generate
       run: |
         echo "Checking that code generation is up-to-date"
-        go install github.com/golang/mock/mockgen@v1.6.0
-        go install github.com/andrewstucki/oapi-codegen/cmd/oapi-codegen@indirection
-        go generate ./...
-        make ctrl-generate
-        make ctrl-manifests
+        make gen
```
sarahalsmiller, line 68: 👏

internal/api/schemas/v1.yaml
```diff
@@ -154,6 +154,103 @@ paths:
...
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/BootstrapToken"
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/Error"
+  "/namespaces/{namespace}/gateways/{name}/tokens/{id}":
```
sarahalsmiller, line 187: Should we have a retrieval endpoint or should we assume that the token is only available on generation? I'm not sure how secure we need these to be.

sarahalsmiller (approved):
Stubs look good to me, just the one question about if we should have the GET endpoint for tokens. We might want to keep it and have it just return some metadata about the tokens, not sure.

internal/api/schemas/v1.yaml
mikemorris: Yea, I wasn't sure about this either, thoughts @andrewstucki?

mikemorris: Will keep this for now and remove later if not needed (assuming all of this may be subject to change depending on how our VM/API work evolves), agreed it might be useful for returning metadata about a token like TTL, revocation status, etc.

PR #386, opened 2022-09-20
Cleanup post-refactor

nathancoleman:
This is the final PR for the breakup of #165 . It just handles a few cleanup items such as removing the `Factory` which is no longer necessary, moving some marshaling funcs and improving the code layout.

### How I've tested this PR:
:robot: tests pass

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/core/resolved.go
```diff
@@ -17,56 +17,66 @@ const (
...
 }
 
-// TODO Is this the best location for this func?
-func UnmarshalRoute(b []byte) (ResolvedRoute, error) {
-	r := &marshaledRoute{}
-	if err := json.Unmarshal(b, r); err != nil {
+// UnmarshalResolvedRoute takes a serialized resolvedRouteWrapper and
+// deserializes it into the appropriate ResolvedRoute implementation.
+// It first partially deserializes to the wrapper in order to get the type
+// and then deserializes the Data field into the target HTTPRoute or TCPRoute.
```
nathancoleman, line 38: @sarahalsmiller @mikemorris we had discussed these functions a bit in #383 . This does seem like the best location IMO. I removed some redundant function wrappers and added godocs in this file to help explain what's going on and why it is this way.

internal/k8s/reconciler/binder_test.go
```diff
@@ -376,9 +375,6 @@ func TestBinder(t *testing.T) {
 			defer ctrl.Finish()
 			client := mocks.NewMockClient(ctrl)
 
-			factory := NewFactory(FactoryConfig{
```
nathancoleman, line 379: `Factory` no longer exists. It was really only passing through a couple of parameters at this point since the gateways and routes continued to be simplified after the factory was introduced in an earlier refactoring

internal/k8s/reconciler/route.go
```diff
@@ -32,10 +30,6 @@ type Route interface {
 type K8sRoute struct {
 	Route
 	RouteState *state.RouteState
-
-	controllerName string
-	logger         hclog.Logger
-	client         gatewayclient.Client
```
nathancoleman, line 38: None of these were actually being used, so the constructor looked more complicated than it needed to be. Removing this is one of things that made it so that the `Factory` is no longer necessary

internal/k8s/reconciler/route_test.go
```diff
@@ -50,36 +42,28 @@ func TestTCPRouteID(t *testing.T) {
...
-	}
-
 	expected := gwv1alpha2.CommonRouteSpec{
 		ParentRefs: []gwv1alpha2.ParentReference{{
 			Name: "expected",
 		}},
 	}
 
-	require.Equal(t, expected, newK8sRoute(&gwv1alpha2.HTTPRoute{
+	assert.Equal(t, expected, newK8sRoute(&gwv1alpha2.HTTPRoute{
```
nathancoleman, line 51: These are just improving our test hygiene. If one case fails, I still want to see that the other cases pass/fail as well instead of having `require` stop the whole test

internal/k8s/service/resolver.go
```diff
@@ -394,15 +393,15 @@ func (r *backendResolver) findCatalogService(service *apigwv1alpha1.MeshService)
...
-	resolved, err := validateCatalogConsulReference(services, service)
+	resolved, err := validateCatalogConsulReference(services)
 	if err != nil {
 		r.logger.Trace("error validating consul services", "error", err)
 		return nil, err
 	}
 	return resolved, nil
 }
 
-func validateCatalogConsulReference(services []*api.CatalogService, object client.Object) (*ResolvedReference, error) {
```
nathancoleman, line 405: The linter pointed out that this parameter wasn't being used

mikemorris (approved):
LGTM

PR #391, opened 2022-09-22
Support distroless (and distroful) Envoy images

nathancoleman:
Remove dependency on shell by:
1. Creating a `Secret` containing the Consul CA cert for each gateway deployment
    1. > **Note** Creating this secret per-gateway is important because the `Secret` must reside in the same namespace as the `Pod` that we will be mounting the volume into, and `Gateways` are not restricted to a single namespace.
    2. > **Note** Creating/updating this `Secret` requires updating [RBAC for the controller over in consul-k8s](https://github.com/hashicorp/consul-k8s/blob/113a266b687881e55676bd8cfa77f3275685368d/charts/consul/templates/api-gateway-controller-clusterrole.yaml#L90-L97). I considered using a `ConfigMap` instead of a `Secret` - which we already conveniently [have permissions to create/update](https://github.com/hashicorp/consul-k8s/blob/113a266b687881e55676bd8cfa77f3275685368d/charts/consul/templates/api-gateway-controller-clusterrole.yaml#L48-L59) - but consul-k8s uses `Secrets` for the ca cert and it feels better to respect that precedent.
2. Mounting a volume on each gateway deployment sourcing the CA cert from the `Secret`
3. Configuring an envar on each gateway deployment w/ the location of the CA cert in the mounted volume
4. Run consul-api-gateway binary directly without the `export` or `exec` that we previously depended on

### How I've tested this PR:
1. Deploy consul-k8s with the API gateway controller enabled (`apiGateway.enabled: true`) and a build of this branch
2. Modify the consul-api-gateway `GatewayClassConfig` to use `envoyproxy/envoy-distroless` instead of `envoyproxy/envoy`
3. Create a `Gateway` and verify that it works correctly

See above

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


internal/k8s/controllers/gateway_controller.go
```diff
@@ -91,6 +91,7 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		Owns(&corev1.ServiceAccount{}).
+		Owns(&corev1.Secret{}).
```
nathancoleman, line 94: TODO The `Secret` doesn't yet get cleaned up when deleting a `Gateway`

.github/workflows/conformance_with_build.yml
```diff
@@ -35,22 +35,22 @@ jobs:
             api-gateway-image: "consul-api-gateway:local-build"
             consul-image: "hashicorp/consul:1.11"
             envoy-image: "envoyproxy/envoy:v1.20-latest"
-            consul-k8s-version: "v0.48.0"
+            consul-k8s-version: "release/0.49.x"
```
nathancoleman, line 38: We can remove this once consul v0.49.0 releases

config/rbac/role.yaml
```diff
@@ -94,8 +94,10 @@ rules:
   resources:
   - secrets
   verbs:
+  - create
```
nathancoleman, line 97: The controller needs to be able to create `Secrets` to hold the Consul CA cert for each gateway `Deployment`. I used a `Secret` to be consistent w/ consul-k8s which uses a `Secret` for the same value.

internal/k8s/builder/gateway.go
```diff
@@ -249,8 +252,13 @@ func (b *GatewayDeploymentBuilder) podSpec() corev1.PodSpec {
 						},
 					},
 				},
+				{
+					Name:  api.HTTPCAFile,
+					Value: consulCALocalFile,
```
nathancoleman, line 257: This adds the path to the CA cert in the pod's environment. This was previously done inside the shell script below as
```shell
export CONSUL_CACERT={{ .ConsulCAFile }}
```

internal/k8s/builder/gateway.go
```diff
@@ -249,8 +252,13 @@ func (b *GatewayDeploymentBuilder) podSpec() corev1.PodSpec {
...
+				},
 			},
-			Command: b.execCommand(),
+			Command: []string{"/bootstrap/consul-api-gateway", "exec"},
+			Args:    b.execArgs(),
```
nathancoleman, line 261: Instead of running the shell with one giant argument that was our startup script, we're now running `consul-api-gateway exec` with many small arguments/flags

internal/k8s/builder/gateway.go
```diff
@@ -319,7 +330,17 @@ func (b *GatewayDeploymentBuilder) volumes() ([]corev1.Volume, []corev1.VolumeMo
...
+					SecretName: b.gateway.Name,
+					Items: []corev1.KeyToPath{
+						{
+							Key:  "consul-ca-cert",
+							Path: consulCAFilename,
+						},
+					},
+					DefaultMode: nil,
+					Optional:    pointer.Bool(false),
+				},
```
nathancoleman, line 343: Instead of creating an empty volume and then having our shell script insert the CA cert, we now source the volume's contents from the `Secret` containing the CA cert, which is created further down in `internal/k8s/reconciler/deployer.go`

internal/k8s/builder/gateway.go
```diff
@@ -370,33 +391,39 @@ type gwContainerCommandData struct {
...
-{{- end}}
-
-exec /bootstrap/consul-api-gateway exec -log-json \
-  -log-level {{ .LogLevel }} \
-  -gateway-host "{{ .GatewayHost }}" \
-  -gateway-name {{ .GatewayName }} \
+// gwContainerArgsTpl is the template for the command arguments executed in the Envoy container.
+// The resulting args are split on \n to obtain a []string for the pod spec's args.
+// Note: Make sure not to leave whitespace at the beginning or end of any line.
+const gwContainerArgsTpl = `
```
nathancoleman, line 397: Left the godoc here to explain since the template looks kinda weird now.

Basically, the flag name and its value are expected to be two separate items in the `args` array for the pod template. It was still easier to compile the list of args using a template like we were before since the other options is a big string builder with a bunch of conditionally-included `fmt.Sprintf`s.

internal/k8s/gatewayclient/gatewayclient.go
```diff
@@ -388,15 +389,19 @@ func (g *gatewayClient) Update(ctx context.Context, obj client.Object) error {
 	return nil
 }
 
-func (g *gatewayClient) CreateOrUpdateDeployment(ctx context.Context, deployment *apps.Deployment, mutators ...func() error) (bool, error) {
-	operation, err := controllerutil.CreateOrUpdate(ctx, g.Client, deployment, func() error {
+func multiMutatorFn(mutators []func() error) func() error {
```
nathancoleman, line 392: abstracting this out into a function wrapper since we needed the same logic in one more place now for `CreateOrUpdateSecret`

internal/k8s/reconciler/deployer.go
```diff
@@ -52,6 +54,10 @@ func (d *GatewayDeployer) Deploy(ctx context.Context, gateway *K8sGateway) error
 		return err
 	}
 
+	if err := d.ensureSecret(ctx, gateway.Gateway); err != nil {
```
nathancoleman, line 57: You can see here that we create the `Secret` containing the CA Cert _before_ we create the `Deployment` that needs to reference the `Secret` for the contents of one of its volumes.

pkg/apis/v1alpha1/types.go
```diff
@@ -177,12 +177,39 @@ type MeshServiceList struct {
...
+	return true
+}
+
 // MergeService merges a gateway service a onto b and returns b, overriding all of
 // the fields that we'd normally set for a service deployment. It does not attempt
 // to change the service type
 func MergeService(a, b *corev1.Service) *corev1.Service {
 	if !compareServices(a, b) {
-		a.Annotations = b.Annotations
+		b.Annotations = a.Annotations
```
nathancoleman, line 212: This seemed like an oversight. Not sure if it's had any material impact on our software or not up until now

pkg/apis/v1alpha1/types.go
```diff
@@ -177,12 +177,39 @@ type MeshServiceList struct {
 	Items []MeshService `json:"items"`
 }
 
+func MergeSecret(a, b *corev1.Secret) *corev1.Secret {
```
nathancoleman, line 180: Ripped off from the `MergeService` logic below. Just need to determine what the final state of the `Secret` should be if we're modifying one that already exists.

nathancoleman (commented):
Personal review

> **Note** Conformance tests won't pass until consul-k8s v0.49.0, which contains the necessary RBAC changes for create/update `Secret`, is released and consumed

.changelog/391.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+Support distroless Envoy images
```
nathancoleman, line 2:
```suggestion
Support distroless Envoy images (with continued support for distroful images)
```

internal/k8s/builder/testdata/tls-cert.deployment.golden.yaml
```diff
@@ -39,27 +39,31 @@ spec:
...
       containers:
-      - command:
-        - /bin/sh
-        - -ec
-        - |2-
-
-          export CONSUL_CACERT=/consul/tls/ca.pem
-          cat <<EOF >/consul/tls/ca.pem
-          CONSUL_CA_MOCKED
-          EOF
```
mikemorris, line 50: This looks a little different from other tests, should any of this be reflected in the new CA cert logic?

mikemorris (commented):
Holding approval until conformance tests are updated to consul-k8s v0.49.0 and passing, but approach LGTM

mikemorris:
Should this update or add a CI test with distroless images as https://github.com/hashicorp/consul-api-gateway/pull/152 did?

nathancoleman:
@mikemorris that gets difficult because consul-k8s doesn't support distroless yet, so we can't just set `imageEnvoy` in our `values.yaml` file. I figured it'd be easier to wait until they do and then add to our matrix. Thoughts?

internal/k8s/builder/testdata/tls-cert.deployment.golden.yaml
nathancoleman: The `CONSUL_CACERT` envvar is now added to the pod spec below in this case, and the cert content ("CONSUL_CA_MOCKED" here) would now be deployed into a secret that the volume mount below references.

I did refine the code a bit just now to only include the envvar and create the secret when the CA is actually required, so now it's clearer that the resulting deployment for this test case which _does_ require the CA is different from the others which don't require the CA.

mikemorris (approved):

PR #393, opened 2022-09-27
Update e2e tests to work on k8s 1.24

nathancoleman:
Kubernetes 1.24 no longer implicitly creates a `Secret` w/ a token in it when a `ServiceAccount` is created. This PR modifies our e2e tests to create the `Secret` explicitly which is the [documented solution](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#to-create-additional-api-tokens).

I found [this](https://itnext.io/big-change-in-k8s-1-24-about-serviceaccounts-and-their-secrets-4b909a4af4e0) to be a helpful explanation of what's happening.

> **Note** A pod running w/ the `ServiceAccount` still has a token mounted inside even though the `Secret` that previously contained this token is no longer created. This is why Consul API Gateway works on K8s 1.24 in a standard deployment; however, our e2e test deployment does not run inside a pod and so requires the changes in this PR.

### How I've tested this PR:
🤖 tests pass

See above

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
 
     - name: Install Dependencies
       run: |
-        curl -L https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64 -o ./kind
+        curl -L https://kind.sigs.k8s.io/dl/v0.16.0/kind-linux-amd64 -o ./kind
         chmod +x ./kind
         mv ./kind /usr/local/bin/kind
         curl -L https://dl.k8s.io/release/v1.22.0/bin/linux/amd64/kubectl -o ./kubectl
--- a/internal/testing/e2e/kind.go
+++ b/internal/testing/e2e/kind.go
@@ -33,7 +33,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.23.6@sha256:b1fa224cc6c7ff32455e0b1fd9cbfd3d3bc87ecaa8fcb06961ed1afb3db0f9ae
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -112,7 +112,7 @@ jobs:
 
     - name: Install Dependencies
       run: |
-        curl -L https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64 -o ./kind
+        curl -L https://kind.sigs.k8s.io/dl/v0.16.0/kind-linux-amd64 -o ./kind
```
nathancoleman, line 115: New version of kind needed to support the 1.24 control plane image

nathancoleman (commented):
Personal review

.github/workflows/conformance_with_build.yml
```diff
@@ -35,22 +35,22 @@ jobs:
             api-gateway-image: "consul-api-gateway:local-build"
             consul-image: "hashicorp/consul:1.11"
             envoy-image: "envoyproxy/envoy:v1.20-latest"
-            consul-k8s-version: "v0.48.0"
+            consul-k8s-version: "release/0.49.x"
```
mikemorris, line 38: Could we wait til the upcoming consul-k8s release ~Thursday to keep this pinned to a released version? IIRC switching to an unreleased version may additionally require specifying an additional `imageK8s` config, see https://github.com/hashicorp/consul-api-gateway/pull/196

nathancoleman: I'm tinkering a little bit atm to determine why conformance is failing on the new control plane image

mikemorris (approved):

PR #397, opened 2022-09-28
Gateway pods fail to start when namespace doesn't exist

sarahalsmiller:
- Duplicated exists functionality from consul-k8s
- Added EnsureNamespace logic to the deployer
- Made minor updates to e2e test to make sure consul enterprise tests are functional with namespace mirroring turned on. 

### How I've tested this PR:
- Spun up a local kind cluster
- Wrote a local e2e test, but that test was failing even when it wasn't hitting the code path, investigating separately

-Tests pass

### Checklist:
- [  X ] Tests added
- [ X ] CHANGELOG entry added 


.changelog/397.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+When a gateway is created in a namespace that doesn't exist in consul, and namespace mirrioring is enabled, the namespace will be created in consul.
```
nathancoleman, line 2:
```suggestion
When a gateway is created in a namespace that doesn't exist in Consul and namespace mirroring is enabled, create the namespace in Consul. 
```

.github/workflows/conformance_with_build.yml
```diff
@@ -35,22 +35,22 @@ jobs:
             api-gateway-image: "consul-api-gateway:local-build"
             consul-image: "hashicorp/consul:1.11"
             envoy-image: "envoyproxy/envoy:v1.20-latest"
-            consul-k8s-version: "v0.49.0"
+            consul-k8s-version: "Consul-api-gateway-create-namespaces"
```
nathancoleman, line 38: This can be removed I imagine since we wound up not needing the RBAC changes we were initially discussing

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+
+const (
+	WildcardNamespace = "*"
+	DefaultNamespace  = "default"
+)
+
+// EnsureNamespaceExists ensures a Consul namespace with name ns exists. If it doesn't,
+// it will create it and set crossNSACLPolicy as a policy default.
+// Boolean return value indicates if the namespace was created by this call.
+func EnsureNamespaceExists(client *capi.Client, ns string, crossNSAClPolicy string) (bool, error) {
```
nathancoleman, line 15:
```suggestion
func EnsureNamespaceExists(client *capi.Client, ns string, crossNSACLPolicy string) (bool, error) {
```

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+	if ns == WildcardNamespace || ns == DefaultNamespace {
```
nathancoleman, line 16: Does the namespace named `default` always exists no matter what?

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return false, err
+	}
+	if namespaceInfo != nil {
+		return false, nil
+	}
+
+	// If not, create it.
+	var aclConfig capi.NamespaceACLConfig
+	if crossNSAClPolicy != "" {
```
nathancoleman, line 31:
```suggestion
	if crossNSACLPolicy != "" {
```

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+		// Create the ACLs config for the cross-Consul-namespace
+		// default policy that needs to be attached
+		aclConfig = capi.NamespaceACLConfig{
+			PolicyDefaults: []capi.ACLLink{
+				{Name: crossNSAClPolicy},
```
nathancoleman, line 36:
```suggestion
				{Name: crossNSACLPolicy},
```

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+			},
+		}
+	}
+
+	consulNamespace := capi.Namespace{
+		Name:        ns,
+		Description: "Auto-generated by consul-k8s",
```
nathancoleman, line 43:
```suggestion
		Description: "Auto-generated by consul-api-gateway",
```

internal/consul/namespaces.go
```diff
@@ -0,0 +1,50 @@
...
+		ACLs:        &aclConfig,
+		Meta:        map[string]string{"external-source": "kubernetes"},
+	}
+
+	_, _, err = client.Namespaces().Create(&consulNamespace, nil)
+	return true, err
```
nathancoleman, line 49: Seems like if `err != nil` then we'd want to return `false, err`

internal/k8s/config.go
```diff
@@ -9,15 +10,17 @@ import (
...
-		SDSPort:  config.SDSServerPort,
-		Logger:   logger,
-		Client:   client,
+		ConsulCA:                  config.CACert,
+		SDSHost:                   config.SDSServerHost,
+		SDSPort:                   config.SDSServerPort,
+		Logger:                    logger,
+		Client:                    client,
+		Consul:                    consulClient,
+		ConsulNamespaceMirrioring: config.ConsulNamespaceConfig.MirrorKubernetesNamespaces,
```
nathancoleman, line 23:
```suggestion
		ConsulNamespaceMirroring: config.ConsulNamespaceConfig.MirrorKubernetesNamespaces,
```

internal/k8s/controller.go
```diff
@@ -155,16 +155,17 @@ func (k *Kubernetes) Start(ctx context.Context) error {
...
+		Client:                    gwClient,
+		Consul:                    k.consul,
+		ConsulCA:                  k.config.CACert,
+		PrimaryDatacenter:         k.config.PrimaryDatacenter,
+		SDSHost:                   k.config.SDSServerHost,
+		SDSPort:                   k.config.SDSServerPort,
+		Logger:                    k.logger.Named("Reconciler"),
+		Store:                     k.store,
+		ConsulNamespaceMapper:     k.config.ConsulNamespaceConfig.Namespace,
+		ConsulNamespaceMirrioring: k.config.ConsulNamespaceConfig.MirrorKubernetesNamespaces,
```
nathancoleman, line 168:
```suggestion
		ConsulNamespaceMirroring: k.config.ConsulNamespaceConfig.MirrorKubernetesNamespaces,
```

internal/k8s/reconciler/deployer.go
```diff
@@ -20,36 +22,49 @@ import (
...
-	primaryDatacenter string
-	sdsHost           string
-	sdsPort           int
+	client                    gatewayclient.Client
+	consulCA                  string
+	primaryDatacenter         string
+	sdsHost                   string
+	sdsPort                   int
+	consul                    *capi.Client
+	consulNamespaceMirrioring bool
```
nathancoleman, line 31:
```suggestion
	consulNamespaceMirroring bool
```

internal/k8s/reconciler/deployer.go
```diff
@@ -20,36 +22,49 @@ import (
...
-	Logger            hclog.Logger
-	Client            gatewayclient.Client
+	ConsulCA                  string
+	PrimaryDatacenter         string
+	SDSHost                   string
+	SDSPort                   int
+	Logger                    hclog.Logger
+	Client                    gatewayclient.Client
+	Consul                    *capi.Client
+	ConsulNamespaceMirrioring bool
```
nathancoleman, line 44:
```suggestion
	ConsulNamespaceMirroring bool
```

internal/k8s/reconciler/deployer.go
```diff
@@ -20,36 +22,49 @@ import (
...
-		sdsPort:           config.SDSPort,
-		logger:            config.Logger,
+		client:                    config.Client,
+		consulCA:                  config.ConsulCA,
+		primaryDatacenter:         config.PrimaryDatacenter,
+		sdsHost:                   config.SDSHost,
+		sdsPort:                   config.SDSPort,
+		logger:                    config.Logger,
+		consul:                    config.Consul,
+		consulNamespaceMirrioring: config.ConsulNamespaceMirrioring,
```
nathancoleman, line 56:
```suggestion
		consulNamespaceMirroring: config.ConsulNamespaceMirroring,
```

internal/consul/namespaces.go
nathancoleman, line 15: It seems like `crossNSACLPolicy` is only ever `""` in our codebase. Do we need to support it anyways, or could we drop the extra parameter for now and add back later if needed?

internal/testing/e2e/consul.go
```diff
@@ -480,7 +480,14 @@ func ConsulHTTPPort(ctx context.Context) int {
 	return mustGetTestEnvironment(ctx).httpPort
 }
 
+func isConsulNamespaceMirrioringOn() bool {
```
nathancoleman, line 483:
```suggestion
func isConsulNamespaceMirroringOn() bool {
```

internal/testing/e2e/consul.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	defaultConsulImage                = "hashicorp/consul:1.11.4"
+	defaultConsulImage                = "hashicorp/consul-enterprise:1.11-ent"
```
nathancoleman, line 31: This means that every time we run e2e tests we need a Consul Enterprise license, right? Is this an intentional change?

nathancoleman (changes requested):
So excited to see this bug patched! Nice work 🎉

internal/consul/namespaces.go
nathancoleman: It seems like `crossNSACLPolicy` is only ever `""` in our codebase. Do we need to support it anyways, or could we drop the extra parameter for now and add back later if needed?

Similarly, we never actually use the `bool` return value. Could we simplify this function to just return the `error`?

internal/testing/e2e/consul.go
sarahalsmiller: Nope that was just for testing on my local. I will say in the CI matrix that value gets overwritten by the env.

internal/consul/namespaces.go
sarahalsmiller: I would assume so, this line is replicated directly from consul-k8s, but I don't know 100%

internal/consul/namespaces.go
sarahalsmiller: I'm ok with dropping crossNSACLPolicy and its logic since I don't think we really want our controller to be in character of NSACL policy to begin with, but I'll keep the boolean for now.

internal/commands/server/k8s_e2e_test.go
```diff
@@ -1036,6 +1036,45 @@ func TestHTTPMeshService(t *testing.T) {
...
 }
 
+func TestGatewayWithConsulNamespaceDoesntExist(t *testing.T) {
+	feature := features.New("gateway class in k8s namespace where consul namespace not created yet, consul enterprise").
+		Assess("gateway is created with appropriate number of replicas set", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+
+			namespace := e2e.Namespace(ctx)
+			client := e2e.ConsulClient(ctx)
+
+			if e2e.IsEnterprise() {
```
nathancoleman, line 1046: Should we only run this test when we're running enterprise? for the oss version, this test isn't really doing anything, right?

I guess it's not really doing any harm, but I could see it being confusing needing to work in this test code one day if it's not clear that the test is only exercised for enterprise.

nathancoleman (approved):

internal/commands/server/k8s_e2e_test.go
sarahalsmiller: I ran into a case where when the code path attempted to run on the non enterprise version, I was getting nil pointer exceptions so I initially let it run for both, but I'm also ok just skipping the test for non enterprise since that path should be covered by the other tests.

nathancoleman (approved):

nathancoleman:
@sarahalsmiller looks like there was a small merge conflict with #405 which just merged

PR #405, opened 2022-10-11
Add max connections configuration to GatewayClassConfig CRD

nathancoleman:
Add a new field to our `GatewayClassConfig` CRD that is piped down to the `ingress-gateway` config entry for overriding the default connection maximum in Envoy (1,024)

### How I've tested this PR:
Create new `Gateways` with and without this setting in the `GatewayClassConfig`. For each `Gateway`, check the corresponding `ingress-gateway` config entry in Consul for the proper override.
<details>
<summary>Example</summary>

```shell
$ kubectl get gatewayclassconfig consul-api-gateway -o yaml | yq .spec.connectionManagement
maxConnections: 4096

$ consul config read -kind ingress-gateway -name api-gateway
{
    "Kind": "ingress-gateway",
    "Name": "api-gateway",
    "TLS": {
        "Enabled": false
    },
    "Listeners": null,
    "Meta": {
        "consul-api-gateway/k8s/Gateway.Name": "api-gateway",
        "consul-api-gateway/k8s/Gateway.Namespace": "consul",
        "external-source": "consul-api-gateway"
    },
    "Defaults": {
        "MaxConnections": 4096,
        "MaxPendingRequests": 0,
        "MaxConcurrentRequests": 0
    },
    "CreateIndex": 42,
    "ModifyIndex": 42
}
```
</details>

See above

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/golang/mock v1.6.0
 	github.com/google/uuid v1.3.0
-	github.com/hashicorp/consul/api v1.15.2
+	github.com/hashicorp/consul/api v1.10.1-0.20221004184615-1b565444beb0
```
nathancoleman, line 16: This PR depends on functionality not yet released. Waiting for `v1.15.3` [here](https://github.com/hashicorp/consul/tags)

nathancoleman: This tag is expected today or tomorrow, so I'm going to go ahead and mark this PR as ready for review

config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
```diff
@@ -36,6 +36,17 @@ spec:
           spec:
             description: Spec defines the desired state of GatewayClassConfig.
             properties:
+              connectionManagement:
+                description: Configuration information for managing connections in
+                  Envoy
+                properties:
+                  maxConnections:
```
nathancoleman, line 43: `connectionManagement.maxConnections` is somewhat long; however, I opted **not** to abbreviate to something like `connMgmt.maxConns` since we haven't abbreviated any other parts of the spec.

internal/core/resolved.go
```diff
@@ -102,7 +102,8 @@ type GatewayID struct {
 }
 
 type ResolvedGateway struct {
-	ID        GatewayID
-	Meta      map[string]string
-	Listeners []ResolvedListener
+	ID             GatewayID
+	Meta           map[string]string
+	Listeners      []ResolvedListener
+	MaxConnections *uint32
```
nathancoleman, line 108: The `ResolvedGateway` is what we have access to when deriving the `ingress-gateway` config entry that we're sending to Consul, so it needs to be aware of any config that we depend on there. This is internal, so it can evolve in the future if we wind up having other connection management config values to pipe through.

internal/k8s/reconciler/gateway.go
```diff
@@ -20,15 +26,15 @@ type K8sGateway struct {
 	*gwv1beta1.Gateway
 	GatewayState *state.GatewayState
 
-	config apigwv1alpha1.GatewayClassConfig
+	Config apigwv1alpha1.GatewayClassConfig
```
nathancoleman, line 29: Exported this so that the JSON marshaler includes it when we serialize for writing to the store backend.

If not exported, we don't have any idea what the max connections value is when rehydrating from the store and syncing into Consul, which we do at an interval in the background for resiliency.

nathancoleman (commented):
Personal review

go.mod
nathancoleman: Tagged and consumed!

sarahalsmiller (approved):
Looks good to me, appreciate your clarifying comments as always

PR #407, opened 2022-10-12
initial scaffolding for consul-server-connection-manager integration

mikemorris:

Integrate the [consul-server-connnection-manager](https://github.com/hashicorp/consul-server-connection-manager) library to replace the reliance on a Consul client when interacting with the Consul API.

Removes a workaround introduced in https://github.com/hashicorp/consul-api-gateway/pull/73 for a bug in Consul 1.11 due to a dependency on the Consul agent for the `ConnectCALeaf` API call - this appears to have been fixed in https://github.com/hashicorp/consul/pull/12820 and was included in the Consul v1.11.6 patch and newer releases.

### How I've tested this PR:

TODO

TBD

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


mikemorris:
Tests seem to mostly be hanging currently, with errors like below when cancelling:
```
failed to resolve DNS name: : lookup : no such host
```
```
failed to resolve DNS name: notadomain: lookup notadomain: no such host
```
If the actual integration is halfway-correct, I'm guessing some mocks might need to be updated.

internal/commands/exec/exec.go
```diff
@@ -206,10 +270,47 @@ func RunExec(config ExecConfig) (ret int) {
...
+		}
+
+		config.ConsulConfig.Transport.TLSClientConfig = tlsClientConfig
+	}
+	config.ConsulConfig.HttpClient.Transport = config.ConsulConfig.Transport
+
+	return api.NewClient(&config.ConsulConfig)
+}
+
+func login(config ExecConfig, s discovery.State) (*api.Client, string, error) {
```
mikemorris, line 305: TODO: login/logout functionality appears to be supported directly within the consul-server-discovery library

internal/consul/certmanager.go
```diff
@@ -236,34 +291,16 @@ func (c *CertManager) Manage(ctx context.Context) error {
 	c.leafWatch.HybridHandler = c.handleLeafWatch
```
mikemorris, line 291: To reimplement this without relying on the Consul agent, we'd basically need to rewrite all of the logic from https://github.com/hashicorp/consul/blob/main/agent/cache-types/connect_ca_leaf.go#L514-L685 to generate a new leaf certificate (which currently depends on internal methods of Consul's `connect` package) and as well as tie it into the CA root watch handler to regenerate certificates on rotation, as the gRPC API only exposes the lower-level `Sign` method.

PR #412, opened 2022-10-13
Update InvalidCertificateRef error to RefNotPermitted

sarahalsmiller:
Closes #396 

- Update InvalidCertificateRef error to RefNotPermitted to match the spec [#1351](https://github-redirect.dependabot.com/kubernetes-sigs/gateway-api/pull/1351)

### How I've tested this PR:
- Update conformance test pin to pull in updated tests

- PR Passes

### Checklist:
- [ X ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,7 @@ jobs:
   e2e:
     name: e2e tests (consul-image=${{ matrix.consul-image }})
     strategy:
+      fail-fast: false
       matrix:
         consul-image:
         - 'hashicorp/consul:1.11.8'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,4 +132,4 @@ jobs:
         --format=short-verbose \
         --jsonfile $TEST_RESULTS_DIR/json/go-test-race.log \
         --junitfile $TEST_RESULTS_DIR/gotestsum-report.xml -- \
-        -tags e2e ./internal/commands/server
+        -tags e2e ./internal/commands/server --failfast
--- a/internal/k8s/reconciler/validator/gateway.go
+++ b/internal/k8s/reconciler/validator/gateway.go
@@ -379,7 +379,6 @@ func (g *GatewayValidator) validateTLS(ctx context.Context, state *state.Listene
 		return err
 	} else if !allowed {
 		nsName := getNamespacedName(ref.Name, ref.Namespace, gateway.Namespace)
-		//g.logger.Warn("Cross-namespace listener certificate not allowed without matching ReferenceGrant", "refName", nsName.Name, "refNamespace", nsName.Namespace)
 		state.Status.ResolvedRefs.RefNotPermitted = rerrors.NewCertificateResolutionErrorNotPermitted(
 			fmt.Sprintf("Cross-namespace listener certificate not allowed without matching ReferenceGrant for Secret %q", nsName))
 		return nil
--- a/internal/k8s/reconciler/validator/gateway.go
+++ b/internal/k8s/reconciler/validator/gateway.go
@@ -380,7 +380,7 @@ func (g *GatewayValidator) validateTLS(ctx context.Context, state *state.Listene
 	} else if !allowed {
 		nsName := getNamespacedName(ref.Name, ref.Namespace, gateway.Namespace)
 		//g.logger.Warn("Cross-namespace listener certificate not allowed without matching ReferenceGrant", "refName", nsName.Name, "refNamespace", nsName.Namespace)
-		state.Status.ResolvedRefs.InvalidCertificateRef = rerrors.NewCertificateResolutionErrorNotPermitted(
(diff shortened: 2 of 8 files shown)


internal/k8s/reconciler/validator/gateway_test.go
```diff
@@ -531,7 +531,7 @@ func TestListenerValidate(t *testing.T) {
 		require.NoError(t, err)
 
 		condition := listenerState.Status.ResolvedRefs.Condition(0)
-		assert.Equal(t, status.ListenerConditionReasonInvalidCertificateRef, condition.Reason)
+		assert.Equal(t, status.ListenerConditionReasonRefNotPermitted, condition.Reason)
```
mikemorris, line 534: If the test name for this is accurate ("unsupported certificate type"), I believe the proper reason here would still be `InvalidCertificateRef`

internal/k8s/reconciler/validator/gateway.go
```diff
@@ -391,7 +391,7 @@ func (g *GatewayValidator) validateTLS(ctx context.Context, state *state.Listene
 		if !errors.As(err, &certificateErr) {
 			return err
 		}
-		state.Status.ResolvedRefs.InvalidCertificateRef = certificateErr
+		state.Status.ResolvedRefs.RefNotPermitted = certificateErr
```
mikemorris, line 394:
```suggestion
		state.Status.ResolvedRefs.InvalidCertificateRef = certificateErr
```
I think reverting this change should fix the expected behavior of the two tests below.

mikemorris (changes requested):
I think reverting one line will fix the expected behavior for two Gateway validator tests, almost there!

internal/k8s/reconciler/validator/gateway_test.go
```diff
@@ -423,7 +423,7 @@ func TestListenerValidate(t *testing.T) {
 		require.NoError(t, err)
 
 		condition := listenerState.Status.ResolvedRefs.Condition(0)
-		require.Equal(t, status.ListenerConditionReasonInvalidCertificateRef, condition.Reason)
+		require.Equal(t, status.ListenerConditionReasonRefNotPermitted, condition.Reason)
```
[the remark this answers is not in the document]
sarahalsmiller: Ok that was something I wasn't super clear on, I'll change it back.

.github/workflows/ci.yml
```diff
@@ -88,6 +88,7 @@ jobs:
   e2e:
     name: e2e tests (consul-image=${{ matrix.consul-image }})
     strategy:
+      fail-fast: false
```
mikemorris, line 91: Is there some set of tests where keeping this is useful alongside `-tags e2e ./internal/commands/server --failfast` below, or was this just for debugging?

mikemorris (approved):

internal/k8s/reconciler/validator/gateway.go
```diff
@@ -380,7 +380,7 @@ func (g *GatewayValidator) validateTLS(ctx context.Context, state *state.Listene
 	} else if !allowed {
 		nsName := getNamespacedName(ref.Name, ref.Namespace, gateway.Namespace)
 		//g.logger.Warn("Cross-namespace listener certificate not allowed without matching ReferenceGrant", "refName", nsName.Name, "refNamespace", nsName.Namespace)
```
nathancoleman, line 382:
```suggestion
```

internal/k8s/reconciler/validator/gateway.go
```diff
@@ -380,7 +380,7 @@ func (g *GatewayValidator) validateTLS(ctx context.Context, state *state.Listene
...
-		state.Status.ResolvedRefs.InvalidCertificateRef = rerrors.NewCertificateResolutionErrorNotPermitted(
+		state.Status.ResolvedRefs.RefNotPermitted = rerrors.NewCertificateResolutionErrorNotPermitted(
```
nathancoleman, line 383: Are we also supposed to return `RefNotPermitted` if the secret referenced does not exist (regardless of ReferenceGrant existence)? Curious if that case is still supposed to use the old error code and, if so, are we respecting that here?

nathancoleman: Ignore the above, you can see us still using `InvalidCertificateRef` in the code below this change when we can't find the Secret

.github/workflows/ci.yml
sarahalsmiller: I would argue that this is useful in general because it keeps us from having to rerun the whole matrix when one test fails, but open to discussion about it.

nathancoleman (approved):

nathancoleman:
Closes #396

.github/workflows/ci.yml
sarahalsmiller: Specifically in the case where the test is flaky and all the other tests already passed.

PR #424, opened 2022-10-21
update to Go v1.19

mikemorris:
- Update to Go v1.19
- Update links in release metadata
- Add Makefile tools to `internal/tools.go` following existing pattern, to allow versioning through `go.mod`
- Update Consul patch releases in CI tests

### How I've tested this PR:
Running all CI tests including conformance.

Verify that updates seem reasonable.

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


nathancoleman:
@mikemorris can you update the pin in `go.mod`? Otherwise LGTM. Exciting update!

mikemorris:
> can you update the pin in go.mod

I was under the impression that should only updated when actually necessary (e.g. introducing usage of a new feature like generics) but I suppose it could be easy to miss that instance if we're only developing/building/testing on Go 1.19 😐 

Thoughts?

nathancoleman:
that makes sense

nathancoleman (approved):

modrake (approved):

PR #433, opened 2022-11-01
Accept PodSecurityPolicy name, bind to managed ServiceAccount if provided

nathancoleman:
> **Note** Depends on the changes proposed in https://github.com/hashicorp/consul-k8s/pull/1672 which was recently merged.

When creating a `Deployment` for a given `Gateway`, we create a "managed" `ServiceAccount` if the `GatewayClassConfig` indicates that we should.

With the changes in this PR, we will also create a `Role` and `RoleBinding` attaching the named `podSecurityPolicy` from the `GatewayClassConfig` if one has been provided.

> **Note** For convenience, the Helm chart changes merged in https://github.com/hashicorp/consul-k8s/pull/1672 will create a `PodSecurityPolicy` that is reasonable for these deployments when `.Values.global.enablePodSecurityPolicies`; however, users may also create their own `PodSecurityPolicy` and reference that in the `GatewayClassConfig` if they wish.

### How I've tested this PR:
Create a K8s cluster that requires `PodSecurityPolicies`. In GKE, this is:

```shell
$ gcloud beta container clusters create cluster-1 --region us-east1 --enable-pod-security-policy
```

Install the Consul Helm chart from https://github.com/hashicorp/consul-k8s/pull/1672 and exercise Consul API Gateway functionality.

See above

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
+++ b/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
@@ -60,7 +60,7 @@ spec:
                     properties:
                       account:
                         description: The name of an existing Kubernetes ServiceAccount
-                          to authenticate as. Ignored if !managed.
+                          to authenticate as. Ignored if managed is true.
                         type: string
                       managed:
                         description: Whether deployments should be run with "managed"
@@ -75,7 +75,7 @@ spec:
                         type: string
                       podSecurityPolicy:
                         description: The name of an existing Kubernetes PodSecurityPolicy
-                          to bind to the ServiceAccount if managed.
+                          to bind to the ServiceAccount if managed is true.
                         type: string
                     type: object
                   ports:
--- a/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
+++ b/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
@@ -60,11 +60,11 @@ spec:
                     properties:
                       account:
                         description: The name of an existing Kubernetes ServiceAccount
-                          to authenticate as. Ignored if !Managed.
+                          to authenticate as. Ignored if !managed.
                         type: string
                       managed:
                         description: Whether deployments should be run with "managed"
-                          service accounts created by the gateway controller.
+                          Kubernetes ServiceAccounts created by the gateway controller.
                         type: boolean
                       method:
                         description: The Consul auth method used for initial authentication
@@ -75,7 +75,7 @@ spec:
                         type: string
                       podSecurityPolicy:
                         description: The name of an existing Kubernetes PodSecurityPolicy
-                          to bind to the ServiceAccount if Managed
+                          to bind to the ServiceAccount if managed.
                         type: string
                     type: object
                   ports:
--- a/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
+++ b/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
@@ -75,7 +75,7 @@ spec:
                         type: string
                       podSecurityPolicy:
                         description: The name of an existing Kubernetes PodSecurityPolicy
-                          to bind to the ServiceAccount if managed is true.
+                          to bind to the managed ServiceAccount if managed is true.
                         type: string
                     type: object
                   ports:
--- a/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
+++ b/config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
@@ -59,8 +59,8 @@ spec:
                     description: Consul authentication information
                     properties:
                       account:
-                        description: The Kubernetes service account to authenticate
-                          as.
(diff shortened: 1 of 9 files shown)


config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
```diff
@@ -73,6 +73,10 @@ spec:
                       namespace:
                         description: The Consul namespace to use for authentication.
                         type: string
+                      podSecurityPolicy:
```
nathancoleman, line 76: I'm not a huge fan of having this named explicitly after `podSecurityPolicy` since it's already going away w/ K8s 1.25; however, the replacement for it is different enough that I don't think it makes sense to try and munge the concepts together into one field. Thoughts?

internal/k8s/controllers/gateway_controller.go
```diff
@@ -47,6 +47,8 @@ type GatewayReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=services,verbs=list;get;create;update;watch
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=list;get;create;watch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=create;update;get;list;watch
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=list;get;create;watch
+//+kubebuilder:rbac:groups=policy,resources=podsecuritypolicies,verbs=use
```
nathancoleman, line 51: The RBAC generated from these annotations is only used for our own testing inside this repo. These changes reflect the changes that have already been made to the Consul Helm chart in https://github.com/hashicorp/consul-k8s/pull/1672

internal/k8s/reconciler/deployer.go
```diff
@@ -87,7 +87,26 @@ func (d *GatewayDeployer) ensureServiceAccount(ctx context.Context, config apigw
...
 
-	return d.client.EnsureServiceAccount(ctx, gateway, serviceAccount)
+	if err := d.client.EnsureServiceAccount(ctx, gateway, serviceAccount); err != nil {
+		return err
+	}
+
+	role := config.RoleFor(gateway)
+	if role == nil {
+		return nil
+	}
```
nathancoleman, line 97: Following the pattern already established by the `ServiceAccount` code immediately above: return `nil` if we shouldn't be creating a `Role` and move along.

pkg/apis/v1alpha1/types.go
```diff
@@ -124,14 +125,16 @@ type CopyAnnotationsSpec struct {
 }
 
 type AuthSpec struct {
-	// Whether deployments should be run with "managed" service accounts created by the gateway controller.
+	// Whether deployments should be run with "managed" Kubernetes ServiceAccounts created by the gateway controller.
 	Managed bool `json:"managed,omitempty"`
 	// The Consul auth method used for initial authentication by consul-api-gateway.
 	Method string `json:"method,omitempty"`
-	// The Kubernetes service account to authenticate as.
+	// The name of an existing Kubernetes ServiceAccount to authenticate as. Ignored if !managed.
```
nathancoleman, line 132:
```suggestion
	// The name of an existing Kubernetes ServiceAccount to authenticate as. Ignored if managed.
```

pkg/apis/v1alpha1/types.go
```diff
@@ -144,17 +147,68 @@ type GatewayClassConfigList struct {
...
+
+	return &rbac.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      gw.Name,
+			Namespace: gw.Namespace,
+			Labels:    utils.LabelsForGateway(gw),
+		},
+		RoleRef: rbac.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
```
nathancoleman, line 189: Looked for some constants for these but had no luck finding any in the K8s SDK

nathancoleman (commented):
Personal review

config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
```diff
@@ -59,12 +59,12 @@ spec:
                     description: Consul authentication information
                     properties:
                       account:
-                        description: The Kubernetes service account to authenticate
-                          as.
+                        description: The name of an existing Kubernetes ServiceAccount
+                          to authenticate as. Ignored if !managed.
```
t-eckert, line 63:
```suggestion
                          to authenticate as. Ignored if managed is true.
```

t-eckert (approved):
Nice! I like that you included some refactoring work as well to keep things tidy

config/crd/bases/api-gateway.consul.hashicorp.com_gatewayclassconfigs.yaml
nathancoleman: integrated your feedback in `types.go` where these CRDs are generated from

nathancoleman: We use the `controller-gen` tool by running `$ make generate`

PR #443, opened 2022-11-09
swap leaf certificate signing from Consul Agent API to gRPC

mikemorris:

Add code extracted from Consul's `connect` package for SPIFFE URI formatting and certificate signing request creation. This code has been isolated such that it should be possible to move into a standalone module at some future point.

TODO:
- [ ] Replace call to `RunWithClientAndHclog` for CA roots watch with gRPC `pbconnectca.WatchRoots`
- [ ] Replace call to `consul.Agent().ConnectCALeaf` with `certificates.GenerateNewLeaf` and `pbconnectca.Sign`
- [ ] Add more unit/integration tests, existing e2e tests should provide sufficient high-level coverage though

### How I've tested this PR:

Tests should be passing on Consul 1.13 and 1.14-dev. Switching to these newer APIs will require dropping support for Consul 1.11 and 1.12 unless we restore the current functionality as a fallback, or find a way to proactively detect the version or feature availability of Consul servers and/or if a Consul agent is available.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


mikemorris:
Why are `internal/consul/exec/exec_test.go` tests apparently expecting gRPC requests to be happening over TLS? The `mockConsulServer` impl doesn't mention TLS anywhere, is there a default I need to unset somewhere?

internal/commands/exec/exec.go
```diff
@@ -156,8 +159,19 @@ func RunExec(config ExecConfig) (ret int) {
 	if config.EnvoyConfig.CertificateDirectory != "" {
 		options.Directory = config.EnvoyConfig.CertificateDirectory
 	}
+
+	tlsConfig, err := api.SetupTLSConfig(&config.ConsulConfig.TLSConfig)
```
mikemorris, line 163: This is wrong, returns an unitialized `tls.Config` struct when TLS is not configured instead of `nil`, which inadvertently attempts client TLS connections when passed to `NewCertManager`

PR #444, opened 2022-11-09
Update `./dev/run` for Kubernetes 1.24

t-eckert:

With Kubernetes 1.24, Service Accounts no longer get a secret with an access token by default.

- Added a step in the `./dev/run` demo that creates the secret. 
- Changed how the secret is accessed in accordance with 1.24's changes. 
- Removed the pinned Consul version from the Helm chart, now we can use what the Consul-k8s Helm chart uses by default.
- Set versions for K8s and Consul-k8s as constants at the top of the file so we can more easily update them in the future.

I chose 1.24 as the K8s version instead of 1.25 because I think Consul-k8s still has some gaps in supporting 1.25.

### How I've tested this PR:

- Ran through [this onboarding doc](https://github.com/hashicorp/consul-api-gateway/blob/main/dev/docs/getting-started.md)

You may also run through the onboarding if you so desire.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/dev/config/helm/consul.yaml
+++ b/dev/config/helm/consul.yaml
@@ -1,6 +1,5 @@
 global:
   name: consul
-  image: "hashicorpdev/consul:581357c32"
   tls:
     enabled: true
     serverAdditionalDNSSANs:
--- a/dev/config/k8s/service-account-secret.yaml
+++ b/dev/config/k8s/service-account-secret.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: consul-api-gateway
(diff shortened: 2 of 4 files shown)


mikemorris (approved):
LGTM, tested locally to confirm the initialization errors below are resolved with this change applied, tweaks to use current (and configurable) versions of k8s and consul-k8s look good too, thanks!
```
Error from server (NotFound): secrets "null" not found
Bootstrapping ACLs
Creating ACL Auth Rule Binding
Missing required '-kubernetes-service-account-jwt' flag
```

PR #446, opened 2022-11-10
Add go-discover binary for exec-based discovery

andrewstucki:

To go along with the movement towards the use of `consul-server-connection-manager`, I noticed that to make the "exec" mode actually useful, `consul-dataplane` vendors the [`hashicorp/go-discover` binary](https://github.com/hashicorp/consul-dataplane/blob/23748fbd2bb740b1a2fecaf58bc5b5a38d67d1d0/Dockerfile#L82). This does the same to make sure we actually have useful "discovery" binaries on disk for any `exec=...` host strings passed to `consul-server-connection-manager`.

Additionally, since we execute in the context of the envoy container rather than our own envoy vendored container (a deliberate decision early on to avoid having to rev our own containers if envoy CVEs dropped), I added a `PATH` environment variable that I believe should be sufficient for executing both `discover` and whatever other binaries someone is likely to package in a custom `envoy` container (in any official release, `envoy` is at `/usr/local/bin/envoy`).

I'll verify this all when we merge the connection-manager work.

Since I think this can be part of a changelog entry regarding `consul-server-connection-manager` incorporation, marking this as `pr/no-changelog`.

### How I've tested this PR:

Unit tests, will verify e2e with actual `exec=...` before merge.

### DIFF
--- a/internal/testing/e2e/consul.go
+++ b/internal/testing/e2e/consul.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/cenkalti/backoff"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
@@ -28,10 +29,12 @@ import (
 )
 
 const (
-	defaultConsulImage                = "hashicorp/consul-enterprise:1.11.4"
+	grpcConsulIncompatibleNameVersion = "1.14"
+	defaultConsulImage                = "hashicorppreview/consul:1.14-dev"
 	envvarConsulImage                 = envvarPrefix + "CONSUL_IMAGE"
 	envvarConsulEnterpriseLicense     = "CONSUL_LICENSE"
 	envvarConsulEnterpriseLicensePath = "CONSUL_LICENSE_PATH"
+	envvarGRPCName                    = "CONSUL_GRPC_VAR_NAME"
 	configTemplateString              = `
 {
 	"log_level": "trace",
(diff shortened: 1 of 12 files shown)


internal/testing/e2e/consul.go
```diff
@@ -253,6 +256,24 @@ func consulCASecret(namespace string, caCert *testing.CertificateInfo) client.Ob
...
+	breakingVersion, err := semver.NewVersion(grpcConsulIncompatibleNameVersion)
+	if err != nil {
+		return "grpc"
+	}
+	// we check major/minor directly since the semver check fails with the trailing -dev tag
+	if breakingVersion.Major() > imageVersion.Major() || (breakingVersion.Major() == imageVersion.Major() && breakingVersion.Minor() > imageVersion.Minor()) {
+		return "grpc"
+	}
+	return "grpc_tls"
+}
```
andrewstucki, line 275: This hack is due to the breaking change introduced in https://github.com/hashicorp/consul/commit/b9a9e1219c67778ff6662a494ed5482c088fe798#diff-d0522c3ebbb7255ee8b82416d7f46bbb05093ff8e25cde2000a46bd77189f7d7R6. Specifically the parameter name for gRPC configuration that's TLS-based changed from just `grpc` to `grpc_tls` and the ports for plaintext and TLS-based connections were split up.

Since we leverage Consul configured to use TLS in our e2e tests, without specifying the parameter in our testing consul config as `grpc_tls` on the latest consul images, then our test gateways wind up trying to use TLS on a plaintext connection and all xDS breaks. We'll need to ensure that this is handled properly on the `consul-k8s` [side of things](https://github.com/hashicorp/consul-k8s/blob/34c5774c2c53f7fca7f0819c7b31bd341dcc72ba/charts/consul/templates/api-gateway-gatewayclassconfig.yaml#L28-L29) since pinning the gateways xDS ports to 8502 when TLS is enabled in the cluster means we'll be breaking things, talking to a plaintext connection with TLS.

sarahalsmiller (approved):
Thanks for catching this

PR #449, opened 2022-11-11
Integrate consul-server-connection-manager library

sarahalsmiller:
### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


internal/commands/exec/exec.go
```diff
@@ -76,28 +81,53 @@ func RunExec(config ExecConfig) (ret int) {
...
+				Namespace:   config.AuthConfig.Method,
+				BearerToken: config.AuthConfig.Token,
+			},
+		},
+	}
+	if config.isTest {
+		discoveryConfig.ServerWatchDisabled = true
+	}
+
+	consulServerConnMgr, err := discovery.NewWatcher(wctx, discoveryConfig, config.Logger)
```
andrewstucki, line 106: So all of the discovery stuff should be moved into the new `consul.Client` interface implementation. You'll need to actually run a watcher goroutine to handle asynchronously swapping out the HTTP client when the returned server addresses change. I'll see if I can whip something up to demonstrate.

andrewstucki: @sarahalsmiller so here's actually pretty close to what I think you'll want, minus figuring out how to initialize all of this stuff 😅 https://github.com/hashicorp/consul-api-gateway/compare/sa-consul-server-connection-manager...consul-server-manager-demo

internal/consul/connection.go
```diff
@@ -1,50 +1,192 @@
...
+	go watcher.Run()
+	defer watcher.Stop()
+
+	// Wait for initial state.
+	state, err := watcher.State()
+	if err != nil {
+		return err
+	}
+	updateClient := func(s discovery.State) error {
+		cfg := api.DefaultConfig()
```
andrewstucki, line 96: Since we do some additional TLS overriding in our entrypoint, I think we'll need to do that here too. We probably want to pass the base client config for Consul in via `ClientConfig` and then just explicitly override the address and token based on the discovery state.

internal/commands/exec/exec.go
```diff
@@ -76,28 +76,7 @@ func RunExec(config ExecConfig) (ret int) {
...
-		consulConfig := &config.ConsulConfig
-		consulConfig.Namespace = config.GatewayConfig.Namespace
-		consulClient, err = api.NewClient(consulConfig)
-		if err != nil {
-			config.Logger.Error("error reinitializing consul client", "error", err)
-			return 1
-		}
-	}
-	client := consul.NewClient(consulClient)
+	client := consul.NewClient(config.ConsulClientConfig)
```
andrewstucki, line 79: I believe you'll need to invoke `client.Watch` here.

internal/consul/connection.go
sarahalsmiller: Do my updates match more what you had in mind?

internal/commands/exec/command.go
```diff
@@ -165,6 +170,37 @@ func (c *Command) Run(args []string) (ret int) {
 		bearerToken = strings.TrimSpace(string(data))
 	}
 
+	addresses, grpcPort, err := c.parseDiscoveryAddresses()
+	if errors.Is(err, discoverySyntaxError) {
```
andrewstucki, line 174: Unless you're seeing issues with passing ips/dns strings to the discovery library, again I don't think we need to do the `exec=` check.

internal/commands/exec/command.go
```diff
@@ -165,6 +170,37 @@ func (c *Command) Run(args []string) (ret int) {
...
+		//old syntax
+		grpcPort = c.flagConsulXDSPort
```
andrewstucki, line 176: This should always be the case, the port that we pass at the end of `CONSUL_HTTP_ADDR` is always the `http` port, not grpc. The way the discover library works is that it hands back ip/dns addresses to use for client initialization, but you have to pass the http port individually.

internal/commands/exec/command.go
```diff
@@ -229,3 +266,18 @@ Usage: consul-api-gateway exec [options]
 	Handles service registration, certificate rotation, and spawning envoy.
 `
 }
+
+func (c *Command) parseDiscoveryAddresses() (addresses string, port int, err error) {
```
andrewstucki, line 270: the conflation of `CONSUL_HTTP_ADDR` really only happens in the server entrypoint since we've always distinguished address v. port for the purposes of our deployments -- so this logic shouldn't be necessary here.

internal/commands/exec/command.go
```diff
@@ -165,6 +170,37 @@ func (c *Command) Run(args []string) (ret int) {
...
+		addresses = c.flagConsulHTTPAddress
+	} else if err != nil {
+		logger.Error("error reading discovery addresses", "error", err)
+		//TODO should this return error?
+		return 1
+	}
+
+	consulClientConfig := consul.ClientConfig{
+		ApiClientConfig: cfg,
+		Addresses:       addresses,
```
andrewstucki, line 186: this should just always be `c.flagConsulHTTPAddress` which may or may not be an `exec=` string.

internal/commands/server/server.go
```diff
@@ -54,12 +55,7 @@ func RunServer(config ServerConfig) int {
 		return 1
 	}
 
-	consulClient, err := api.NewClient(config.ConsulConfig)
-	if err != nil {
-		config.Logger.Error("error creating a Consul API client", "error", err)
-		return 1
-	}
-	client := consul.NewClient(consulClient)
+	client := consul.NewClient(config.ConsulClientConfig)
```
andrewstucki, line 58: this is where you'll want to do all of the splitting logic

internal/commands/server/server.go
```diff
@@ -24,13 +24,14 @@ import (
...
-	Logger            hclog.Logger
-	ConsulConfig      *api.Config
-	K8sConfig         *k8s.Config
-	ProfilingPort     int
-	MetricsPort       int
-	PrimaryDatacenter string
+	Context            context.Context
+	Logger             hclog.Logger
+	ConsulConfig       *api.Config
+	ConsulClientConfig consul.ClientConfig
```
andrewstucki, line 30: It doesn't look like this is initialized anywhere.

andrewstucki (commented):
More comments, maybe we can pair tomorrow on this.

andrewstucki:
So I went ahead and validated this with the following agentless configurations that pretty much exercise the majority auth method + tls configuration for the discovery library:

internal server
TLS enabled + internal server
Managed ACLs + internal server
External server with static address
External server with connection string `exec=echo 192.168.65.2`

Despite the introduction of all the more flaky unit tests due to the upstream issues with the manager library use of the grpc `balancer` subpackage, I think we can merge with the inclusion of some changelog entries + a thorough review pass. We'll continue to vet this over the next few days with a variety of consul-k8s configurations. @sarahalsmiller since I'm out a good chunk of tomorrow morning, would you mind handling the changelog stuff?

internal/commands/exec/exec_test.go
```diff
@@ -32,22 +49,21 @@ func TestRunExecLoginError(t *testing.T) {
...
 			Token:  "token",
 		},
 		isTest: true,
 	}))
-	require.Contains(t, buffer.String(), "error logging into consul")
+	require.Contains(t, buffer.String(), "no such file or directory")
 }
 
 func TestRunExecLoginSuccessRegistrationFail(t *testing.T) {
-	t.Parallel()
```
mikemorris, line 49: I'm assuming this was needed because of the un-threadsafe behavior of how consul-server-connection-manager is registering the gRPC load balancer? Refs https://github.com/hashicorp/consul-server-connection-manager/pull/25

internal/consul/connection.go
```diff
@@ -11,40 +32,183 @@ type Client interface {
...
+
+	config := discovery.Config{
+		Addresses:   c.config.Addresses,
+		GRPCPort:    c.config.GRPCPort,
+		Credentials: c.config.Credentials,
+	}
+
+	if !c.config.PlainText {
+		config.TLS = c.config.TLS
+	}
```
mikemorris, line 105: I was curious about the best way to check for whether TLS should be used for these gRPC connections - this approach seems reasonable, in 
I took a different approach more similar to `internal/commands/server/server.go` and `internal/commands/exec/exec.go` though, checking
```
if consulAPITLSConfig.CAFile != "" || len(consulAPITLSConfig.CAPem) > 0
```
It looks like consul-dataplane is asking for an explicit `-tls-disabled` configuration https://github.com/hashicorp/consul-dataplane/#configuration-reference

internal/testing/e2e/consul.go
```diff
@@ -531,12 +534,12 @@ func CreateConsulACLPolicy(ctx context.Context, cfg *envconf.Config) (context.Co
 		return ctx, nil
 	}
 	env := consulEnvironment.(*consulTestEnvironment)
-	token, _, err := env.consulClient.ACL().Bootstrap()
+	token, _, err := env.consulClient.Internal().ACL().Bootstrap()
```
mikemorris, line 537: Could we add a `ACL()` method to the client wrapper since we plan to remove the `Internal()` catch-all soon?

internal/testing/e2e/gateway.go
```diff
@@ -40,7 +40,7 @@ func (p *gatewayTestEnvironment) run(ctx context.Context, namespace string, cfg
 	consulClient := ConsulClient(ctx)
 
 	// this should go away once we implement auth in the server bootup
-	consulClient.AddHeader("x-consul-token", ConsulInitialManagementToken(ctx))
+	consulClient.Internal().AddHeader("x-consul-token", ConsulInitialManagementToken(ctx))
```
mikemorris, line 43: Is this still needed at all now that we're using the consul-server-connection-manager integrated login flow?

internal/commands/exec/exec_test.go
```diff
@@ -3,27 +3,44 @@ package exec
...
+	"github.com/hashicorp/consul/proto-public/pbdataplane"
 	"github.com/hashicorp/go-hclog"
 
+	"github.com/hashicorp/consul-api-gateway/internal/consul"
 	gwTesting "github.com/hashicorp/consul-api-gateway/internal/testing"
 )
 
-func TestRunExecLoginError(t *testing.T) {
-	t.Parallel()
+type testDPServer struct{}
```
mikemorris, line 31: Do we not actually need to mock the [`WatchServers`](https://github.com/hashicorp/consul/blob/main/proto-public/pbserverdiscovery/serverdiscovery.proto#L13) method anywhere? Are we only exercising that in e2e tests?

mikemorris (approved):
Minor notes/questions but mostly LGTM!

internal/testing/e2e/consul.go
sarahalsmiller, line 537:
```suggestion
	token, _, err := env.consulClient.ACL().Bootstrap()
```

internal/commands/exec/exec_test.go
andrewstucki: No, we don't -- this is just a gRPC server that stubs out the endpoints. So the library actually hits this and issues real API calls against this. It's almost like using `httptest`.

internal/testing/e2e/gateway.go
andrewstucki: Yes, this is the client that gets passed around with the root auth token in our tests so that we can actually check what's in Consul.

internal/commands/exec/exec_test.go
andrewstucki: 👍

internal/commands/exec/command.go
```diff
@@ -151,11 +151,6 @@ func (c *Command) Run(args []string) (ret int) {
 	if cfg.TLSConfig.CAFile != "" {
 		cfg.Scheme = "https"
 	}
-	consulClient, err := api.NewClient(cfg)
-	if err != nil {
-		logger.Error("error creating consul client", "error", err)
-		return 1
-	}
```
andrewstucki, line 156: The call to `api.NewClient` actually mutates the config under the hood -- for now let's just keep this and ignore the returned client, so:

```go
	// this call mutates the cfg object with a bunch of defaults
	// so we're going to keep it for now
	if _, err := api.NewClient(cfg); err != nil {
		logger.Error("error creating consul client", "error", err)
		return 1
	}
```

andrewstucki (changes requested):

internal/commands/exec/command.go
mikemorris: Ugh that's an unfortunate behavior.

PR #450, opened 2022-11-15
Add TLS server name and partition CRD extensions

andrewstucki:

This adds support for setting the `CONSUL_PARTITION` and `CONSUL_TLS_SERVER_NAME` environment variables to support some of our changes for agentless.

### How I've tested this PR:

Installed via helm and saw the values being set and the container starting up even when the values are empty.

### Checklist:
- [x] CHANGELOG entry added

### DIFF
--- a/pkg/apis/v1alpha1/types.go
+++ b/pkg/apis/v1alpha1/types.go
@@ -95,7 +95,8 @@ type ConsulSpec struct {
 	// The scheme to use for connecting to Consul.
 	// +kubebuilder:validation:Enum=http;https
 	Scheme string `json:"scheme,omitempty"`
-	// The partition in which the gateway is registered.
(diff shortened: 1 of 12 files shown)


pkg/apis/v1alpha1/types.go
```diff
@@ -95,6 +95,12 @@ type ConsulSpec struct {
 	// The scheme to use for connecting to Consul.
 	// +kubebuilder:validation:Enum=http;https
 	Scheme string `json:"scheme,omitempty"`
+	// The partition in which the gateway is registered.
```
t-eckert, line 98:
```suggestion
	// The Consul admin partition in which the gateway is registered.
	// https://developer.hashicorp.com/consul/tutorials/enterprise/consul-admin-partitions
```

PR #451, opened 2022-11-15
Update conformance.yml

sarahalsmiller:
### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


.github/workflows/conformance_with_build.yml
```diff
@@ -42,12 +42,12 @@ jobs:
             envoy-image: "envoyproxy/envoy:v1.22-latest"
             consul-k8s-version: "v0.49.0"
```
andrewstucki, line 43: is there a way of running this off of the master branch pre-release?

sarahalsmiller: Yeah I'd just have to add it as a trigger

andrewstucki: 👍 I think doing `consul-k8s` at master would be good for ensuring that we're good in the agentless configuration pre-release.

sarahalsmiller: Actually I'm wrong, we can manually trigger it from main whenever we want already

Screen Shot 2022-11-16 at 2 10 58 PM

sarahalsmiller: Still debugging why consul 1.14 isn't installing

andrewstucki: @sarahalsmiller Ah, I see, sorry, I meant to use the upcoming `consul-k8s` 1.0.0 release (i.e. their main branch) rather than 0.49.0 to install. The helm chart has changed _significantly_ for the agentless work.

sarahalsmiller: Ah I understand! That's still possible, I just need to update the ref value

PR #454, opened 2022-11-17
Add fallback to no gRPC

andrewstucki:

Add an opt-in value for trying to use the new dynamic server discovery components. This is needed because agent (non-server) nodes don't have the requisite gRPC endpoints to actually use the connection manager library. When the `CONSUL_DYNAMIC_SERVER_DISCOVERY` is not set, then we do things in the "old way" without gRPC, otherwise if it's set we know we're talking to a server and can use gRPC.

### How I've tested this PR:

Manually w/ Consul 1.12 + 1.13 and consul-k8s 0.49.1 and consul-k8s main branch in both agent and agentless mode pending my PR that I'll link here.

### DIFF
--- a/internal/commands/server/command.go
+++ b/internal/commands/server/command.go
@@ -172,10 +172,14 @@ func (c *Command) Run(args []string) int {
 		MirrorKubernetesNamespacePrefix: c.flagMirrorK8SNamespacePrefix,
 	}
 
-	consulHTTPAddressOrCommand, port, err := parseConsulHTTPAddress()
+	consulScheme, consulHTTPAddressOrCommand, port, err := parseConsulHTTPAddress()
 	if err != nil {
 		logger.Error("error reading "+consulHTTPAddressEnvName, "error", err)
 		return 1
+	}
+	if consulCfg.Scheme != "https" {
+		// override only if it needs to be explicitly marked
+		consulCfg.Scheme = consulScheme
 	}
 
 	tlsCfg, err := api.SetupTLSConfig(&consulCfg.TLSConfig)
@@ -201,8 +205,10 @@ func (c *Command) Run(args []string) int {
 	}
 
 	consulClientConfig := consul.ClientConfig{
+		Name:            "api-gateway-controller",
 		ApiClientConfig: consulCfg,
 		Addresses:       consulHTTPAddressOrCommand,
+		UseDynamic:      os.Getenv("CONSUL_DYNAMIC_SERVER_DISCOVERY") == "true",
 		HTTPPort:        port,
 		GRPCPort:        grpcPort(),
 		PlainText:       consulCfg.Scheme == "http",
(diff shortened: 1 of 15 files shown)


sarahalsmiller (approved):
thanks for hopping on his andrew. Some stuff we can probably clean up later, but when the tests pass I'm good to merge this.

sarahalsmiller:
adding the conformance pr label just to see if it will pass, but we can remove if its blocking

internal/consul/connection.go
```diff
@@ -81,6 +83,46 @@ func (c *client) Wait(until time.Duration) error {
 }
 
 func (c *client) WatchServers(ctx context.Context) error {
+	if !c.config.UseDynamic {
+		cfg := c.config.ApiClientConfig
+		cfg.Address = fmt.Sprintf("%s:%d", c.config.Addresses, c.config.HTTPPort)
```
mikemorris, line 88: Not entirely specific to this line, but do we have unit test coverage for checking that both `http://` prefixed and plain IPs are handled for agentful and agentless paths?

internal/consul/connection.go
```diff
@@ -81,6 +83,46 @@ func (c *client) Wait(until time.Duration) error {
...
+			client, err = api.NewClient(cfg)
+			if err != nil {
+				c.initialized <- err
+				return err
+			}
+		}
+
+		c.mutex.Lock()
+		c.client = client
+		c.token = cfg.Token
```
mikemorris, line 117: ⚠️  I think this should get set to `token` (and set `token = c.config.Credentials.Static.Token; cfg.Token = token` in the non-login path) unless `login` is mutating `cfg.Token`?

**Edit:** reading the `login` impl it looks like it does mutate `cfg.Token` so this should be okay, but it is a bit unintuitive.

mikemorris (approved):
Minor notes, and could use a changelog entry.

internal/commands/server/command.go
```diff
@@ -172,11 +172,15 @@ func (c *Command) Run(args []string) int {
...
-	consulHTTPAddressOrCommand, port, err := parseConsulHTTPAddress()
+	consulScheme, consulHTTPAddressOrCommand, port, err := parseConsulHTTPAddress()
 	if err != nil {
 		logger.Error("error reading "+consulHTTPAddressEnvName, "error", err)
 		return 1
 	}
+	if consulCfg.Scheme != "https" {
+		// override only if it needs to be explicitly marked
+		consulCfg.Scheme = consulScheme
+	}
```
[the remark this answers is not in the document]
andrewstucki: sorry, this is to say, if we already have the scheme set to `https` (via things like ENV variables, which `DefaultConfig` will source), then don't override it -- otherwise  set it based off of what we got from parsing it out of the `CONSUL_ADDR` (in 1.0.0 we don't set the scheme and use `CONSUL_USE_SSL`, in <= 0.49 we use the scheme as part of the address, this handles both cases)

andrewstucki:
Merging, will take a pass at cleaning some of this up in a few days once we decide how long we need to support this auth flow code.

PR #459, opened 2022-11-18
Add fix for when we need to use the system-wide trusted CAs

andrewstucki:

This adds the simplest path for supporting the helm option `externalServer.useSystemRoots`. Since the controller always passes in a value to the deployment via `CONSUL_CACERT` when the controller knows that the Consul connections are served over TLS, we leverage the fact that it's going to write an empty file to disk that the deployment will then try and use as a file.

Rather than doing a larger re-write of the deployment code to handle cases in which TLS is configured but no certs are passed in (when you want to validate with system-wide certs), I added some code that handles the invalid certificates being passed in, in which case we just fallback to using the system-wide certs. What this entails is:

1. Ensuring that the file passed in `CONSUL_CACERT` is valid.
2. If it isn't, unsetting it but flagging that `CONSUL_USE_SSL` is set to mark that this still needs to talk over SSL
3. Passing this info down to our envoy process manager so that it knows to bootstrap envoy with the proper paths for verifying the Consul SSL connection (i.e. the path to the envoy container's system roots)

### How I've tested this PR:

Verified this on HCP with `useSystemRoots`, with a local consul server node and the self-signed certs that go with it, and with a local node and TLS disabled. These were validated with the changes in https://github.com/hashicorp/consul-k8s/pull/1743. 

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/459.txt
+++ b/.changelog/459.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix being able to use system-wide root certificates in deployments.
+```
--- a/internal/commands/exec/command.go
+++ b/internal/commands/exec/command.go
@@ -2,6 +2,7 @@ package exec
 
 import (
 	"context"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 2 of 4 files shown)


t-eckert (approved):
This is a great workaround!

PR #483, opened 2023-01-04
k8s/controller: always map "default" Consul namespace to empty string

mikemorris:

Always map the `"default"` Consul namespace to the empty string in the memory backend store, for compatibility with Consul OSS and the logic to [parse the namespace of a Gateway from the SPIFFE URI](https://github.com/hashicorp/consul-api-gateway/blob/v0.5.1/internal/envoy/middleware.go#L112-L114), used when [reading a Gateway from a store backend](https://github.com/hashicorp/consul-api-gateway/blob/v0.5.1/internal/envoy/middleware.go#L79-L85).

This Consul namespace mapper function is already called when [reconciling a Gateway to a backend store](https://github.com/hashicorp/consul-api-gateway/blob/v0.5.1/internal/k8s/reconciler/manager.go#L180-L192), but the corresponding lookup fails with Consul Enterprise when `--consul-destination-namespace` is set to `"default"` because of the special case handling in the `parseURI` helper.

No migration handling should be needed because this is all within the ephemeral memory store for now, but this does need an e2e test for a service in the `"default"` namespace on Consul Enterprise.

As a temporary workaround, this edge case can be avoided by configuring `consulNamespaces.mirroringK8S: true` or specfiying `""` or _any_ value other than `"default"` for `consulNamespaces.consulDestinationNamespace` in the Helm chart.

### Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/internal/commands/server/k8s_e2e_test.go
+++ b/internal/commands/server/k8s_e2e_test.go
@@ -169,6 +169,10 @@ func TestGatewayWithoutNamespaceMirroring(t *testing.T) {
 			assert.NoError(t, resources.Delete(ctx, secondGateway))
 			assert.NoError(t, resources.Delete(ctx, thirdGateway))
 
+			// Re-enable namespace mirroring
+			ctx, err := e2e.SetNamespaceMirroring(true)(ctx, nil)
+			require.NoError(t, err)
+
 			return ctx
 		})
 
@@ -344,6 +348,10 @@ func TestGatewayBasic(t *testing.T) {
 
 			// check for the service being registered
 			client := e2e.ConsulClient(ctx)
+            t.Log("k8s namespace:", e2e.Namespace(ctx)
+            t.Log("consul namespace:", e2e.ConsulNamespace(ctx)
+            t.Log("mirroring:", e2e.NamespaceMirroring(ctx)
+
 			require.Eventually(t, func() bool {
 				services, _, err := client.Catalog().Service(gatewayName, "", &api.QueryOptions{
 					Namespace: e2e.ConsulNamespace(ctx),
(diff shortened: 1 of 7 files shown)


internal/commands/server/k8s_e2e_test.go
```diff
@@ -116,6 +116,69 @@ func TestGatewayWithClassConfigChange(t *testing.T) {
...
+			checkGatewayConfigAnnotation(ctx, t, resources, thirdGatewayName, namespace, gcc)
+
+			// Cleanup
+			assert.NoError(t, resources.Delete(ctx, firstGateway))
+			assert.NoError(t, resources.Delete(ctx, secondGateway))
+			assert.NoError(t, resources.Delete(ctx, thirdGateway))
+
+			// Re-enable namespace mirroring
+			ctx, err = e2e.SetNamespaceMirroring(true)(ctx, nil)
+			require.NoError(t, err)
```
mikemorris, line 174: I was surprised this would be necessary, as I expected SetUpStack to create a new context for each test, but it appears to be reused - I'm not sure if that's intentional/common, or if that should be changed and this cleanup could be removed?

sarahalsmiller: I've run into this issue before when touching the e2e tests as well. I'm not sure what the right solution is because I definitely think spinning up a new kind cluster for each test would make these tests take way longer, but the current set up make it difficult to do any kind of namespace manipulation.

sarahalsmiller (approved):
Thanks for hopping on this

andrewstucki (approved):
The patch itself looks fine to me

PR #484, opened 2023-01-18
deps: update consul-server-connection-manager

mikemorris:
This resolves test flakiness issues caused by the un-threadsafety of registering a gRPC balancer in `NewWatcher` rather than an `init()` function. This didn't cause problems in actual deployments, but could happen when in our test suite when multiple calls to `NewWatcher` happened concurrently.

- Updates to a new version of consul-server-connection-manager which removes the gRPC custom balancer implementation and switches to the default [`pick_first`](https://github.com/grpc/grpc/blob/master/doc/load-balancing.md#pick_first) balancer.
- Removes the now-unnecessary `globalWatcherMutex` (added in https://github.com/hashicorp/consul-api-gateway/pull/449)
- Removes an unneeded context cancellation call from the `client.WatchServers` method which was responsible for a race condition occurring infrequently on timeouts, particularly from the [discovery-error](https://github.com/hashicorp/consul-api-gateway/blob/874f3248ad6636d3d1cf995ed16d0706a0fd6c82/internal/commands/exec/command_test.go#L119-L128) unit test.
  - This context cancellation was already handled by the caller at https://github.com/hashicorp/consul-api-gateway/blob/874f3248ad6636d3d1cf995ed16d0706a0fd6c82/internal/commands/exec/exec.go#L85

### How I've tested this PR:

Research and discussion on upstream implementation with @kisunji and @pglass, testing https://github.com/hashicorp/consul-server-connection-manager/pull/28 as a dependency locally and in CI. Thanks for the help in tracking down these issues!

The CI unit tests should stop being flaky.

### Checklist:
- [ ] ~~Tests added~~

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/golangci/golangci-lint v1.50.1
 	github.com/google/uuid v1.3.0
-	github.com/hashicorp/consul-server-connection-manager v0.1.1-0.20221115222743-242744ae73fb
+	github.com/hashicorp/consul-server-connection-manager v0.1.1
 	github.com/hashicorp/consul/api v1.18.0
 	github.com/hashicorp/consul/proto-public v0.2.1
 	github.com/hashicorp/consul/sdk v0.13.0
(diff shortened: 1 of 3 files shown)


nathancoleman (approved):
Thanks for knocking this out!

PR #489, opened 2023-01-30
consul: set Envoy xDS address from discovery state

mikemorris:

Use the discovery state from consul-server-connection-manager to set the xDS address for Envoy gateway instances rather than passing through the `--consul-http-address` CLI argument.

This is intended to help mitigate an issue where Consul API Gateway will hit the Consul gRPC rate limiter when registering a gateway directly with a Consul server due to consul-k8s v1.0 or later running in "agentless" mode.

### How I've tested this PR:

It's failing e2e tests currently, as CSCM seems to not be resolving `host.docker.internal` to a Consul server IP successfully, despite the preceding call to `Wait()`.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/commands/exec/command.go
+++ b/internal/commands/exec/command.go
@@ -220,7 +220,6 @@ func (c *Command) Run(args []string) (ret int) {
 		},
 		EnvoyConfig: EnvoyConfig{
 			CACertificateFile: cfg.TLSConfig.CAFile,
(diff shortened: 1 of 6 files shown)


sarahalsmiller:
Initial fix didn't work, more investigation required

PR #504, opened 2023-02-09
QoL: Apply make fmt

jm96441n:
ran `make fmt` so everything is properly fmt'd now

### How I've tested this PR:
I ran the tests

Code Review

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,6 @@ jobs:
       run: |
         echo "Checking that code generation is up-to-date"
         make gen
-        make fmt
         git diff --quiet || (echo 'Working tree is dirty' && git --no-pager diff && exit 1)
 
     - name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
       run: |
         echo "Checking that code generation is up-to-date"
         make gen
+        make fmt
         git diff --quiet || (echo 'Working tree is dirty' && git --no-pager diff && exit 1)
 
     - name: Test
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,12 @@ endif
 test:
 	go test ./...
 
+# Run e2e tests, takes an optional variable `consul_image` which specifies the consul image to test against
+.PHONY: e2e
+e2e:
+	E2E_APIGW_CONSUL_IMAGE=$(consul_image) ./scripts/e2e_local.sh
+
+
 generate-golden-files:
 	GENERATE=true go test ./internal/adapters/consul
 	GENERATE=true go test ./internal/envoy
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ GOIMPORTS=$(shell which goimports)
 
 .PHONY: fmt
 fmt: goimports
-	@for d in $$(go list -f {{.Dir}} ./...); do ${GOIMPORTS} --local github.com/hashicorp --local github.com/hashicorp/consul-api-gateway -w -l $$d/*.go; done
+	@for d in $$(go list -f {{.Dir}} ./...); do if [[ $$d != *"mocks"* ]]; then ${GOIMPORTS} --local github.com/hashicorp/consul-api-gateway,github.com/hashicorp -w -l $$d/*.go; fi; done
 
 .PHONY: lint
 lint:
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ generate-golden-files:
 	GENERATE=true go test ./internal/k8s/builder
 
 .PHONY: gen
-gen: generate-golden-files ctrl-generate ctrl-manifests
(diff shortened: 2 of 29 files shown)


internal/commands/exec/command.go
```diff
@@ -15,15 +15,17 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/consul-api-gateway/internal/consul"
 	"github.com/hashicorp/consul-server-connection-manager/discovery"
 
+	"github.com/hashicorp/consul-api-gateway/internal/consul"
```
nathancoleman, line 20: Should be further down w/ the other local pkg

jm96441n: oh weird, lemme check out our implementation of `make fmt` to make sure it's not missing anything

Makefile
```diff
@@ -37,7 +37,7 @@ GOIMPORTS=$(shell which goimports)
 
 .PHONY: fmt
 fmt: goimports
-	@for d in $$(go list -f {{.Dir}} ./...); do ${GOIMPORTS} --local github.com/hashicorp --local github.com/hashicorp/consul-api-gateway -w -l $$d/*.go; done
+	@for d in $$(go list -f {{.Dir}} ./...); do ${GOIMPORTS} --local github.com/hashicorp/consul-api-gateway,github.com/hashicorp -w -l $$d/*.go; done
```
jm96441n, line 40: made this change, based on the output of `goimports -h` the `--local` flag expects a comma separated list:
```
 -local string
        put imports beginning with this string after 3rd-party packages; comma-separated list
```

internal/commands/server/command.go
```diff
@@ -14,14 +14,16 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/consul-api-gateway/internal/consul"
 	"github.com/hashicorp/consul-server-connection-manager/discovery"
 
-	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/consul-api-gateway/internal/consul"
```
nathancoleman, line 19: hmmm I would expect this to be in the last group of imports down there ⬇️

nathancoleman (changes requested):

internal/commands/server/command.go
jm96441n: lemme move this manually, `goimports` is being kinda funky and just leaving things if they're already in a separate group

Makefile
```diff
@@ -37,7 +37,7 @@ GOIMPORTS=$(shell which goimports)
 
 .PHONY: fmt
 fmt: goimports
-	@for d in $$(go list -f {{.Dir}} ./...); do ${GOIMPORTS} --local github.com/hashicorp --local github.com/hashicorp/consul-api-gateway -w -l $$d/*.go; done
+	@for d in $$(go list -f {{.Dir}} ./...); do if [[ $$d != *"mocks"* ]]; then ${GOIMPORTS} --local github.com/hashicorp/consul-api-gateway,github.com/hashicorp -w -l $$d/*.go; fi; done
```
jm96441n, line 40: so this is mostly working, we get a list of directories passed in which allows us to exclude anything in a `mocks` directory, excluding the other generated files like `pkg/apis/v1alpha1/zz_generated.deepcopy.go` is going to be a little more tricky, I was thinking of doing something hacky like running `git checkout $(git rev-parse --abbrev-ref HEAD) "*generated*"` after the fmt command completes, but that feels icky to me

jm96441n: so we can actually get around this just diffing against fmt'd files in CI

jm96441n: https://github.com/hashicorp/consul-api-gateway/pull/504/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR69

.github/workflows/ci.yml
```diff
@@ -66,6 +66,7 @@ jobs:
       run: |
         echo "Checking that code generation is up-to-date"
         make gen
+        make fmt
```
jm96441n, line 69: would it be beneficial here to add an additional CI check to ensure the code has been `fmt'd` so we keep the repo in a clean state?

mikemorris: Possibly, although I think the `lint` check already covers this because of https://github.com/hashicorp/consul-api-gateway/blob/main/.golangci.yml#L23-L25

Makefile
mikemorris: 🤔 Hmm, I think we might want to move ^ into the `make gen` task itself to avoid a false local diff if you forget to run `fmt` afterwards?

jm96441n: oh good call, let me add that there as well

jm96441n: and that actually should eliminate the need to run `fmt` in the ci pipeline because the command already handles it

mikemorris (approved):

PR #505, opened 2023-02-10
Fix bug with cross namespace policy

### LINKED ISSUE #510: API Gateway doesn't attach cross-namespace lookup policy on namespace creation
### Overview of the Issue

This issue impacts API Gateway instances created in Kubernetes, with Consul Enterprise clusters leveraging namespace mirroring.

When an API Gateway instance is launched in a Kubernetes namespace whose equivalent does not yet exist in Consul, it creates a matching namespace automatically. This mimics the behavior of consul connect-inject in the same scenario.

The created namespace needs a default policy assigned permitting cross-namespace service lookup, which is currently skipped by API Gateway Controller. Without this policy, the ACL token given to the API Gateway pods cannot discover the appropriate endpoints for any target services that are not in the same namespace as the gateway.

### Reproduction Steps

* Deploy Consul Enterprise to a kubernetes cluster with the following features enabled:
```yaml
global:
  image: "hashicorp/consul-enterprise:1.14.3-ent"
  enableConsulNamespaces: true
  acls:
    manageSystemACLs: true
connectInject:
  enabled: true
  consulNamespaces:
    mirroringK8S: true
apiGateway:
  enabled: true
  image: hashicorp/consul-api-gateway:0.5.1
  imageEnvoy: envoyproxy/envoy:v1.23.1
```

* Create a new kubernetes namespace, and deploy a Gateway resource to that namespace:
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: apigw
  namespace: apigateway
spec:
  gatewayClassName: consul-api-gateway
  listeners:
  - protocol: HTTP
    port: 8080
    name: http
    allowedRoutes:
      namespaces:
        from: All
```

* Create a different kubernetes namespace, and deploy a reference connect-injected application and service:
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: dashboard
  namespace: demoapp
---
apiVersion: v1
kind: Service
metadata:
  name: dashboard
  namespace: demoapp
spec:
  selector:
    app: dashboard
  ports:
  - port: 9002
    targetPort: 9002
    name: dashboard
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: dashboard
  name: dashboard
  namespace: demoapp
spec:
  replicas: 1
  selector:
    matchLabels:
      app: dashboard
  template:
    metadata:
      annotations:
        'consul.hashicorp.com/connect-inject': 'true'
        'consul.hashicorp.com/connect-service-upstreams': 'counting:9001'
      labels:
        app: dashboard
    spec:
      serviceAccountName: dashboard
      containers:
      - name: dashboard
        image: hashicorp/dashboard-service:0.0.4
        ports:
        - containerPort: 9002
        env:
        - name: COUNTING_SERVICE_URL
          value: 'http://127.0.0.1:9001'
---
apiVersion: consul.hashicorp.com/v1alpha1
kind: ServiceDefaults
metadata:
  name: dashboard
  namespace: demoapp
spec:
  protocol: http
```

* Define an HTTPRoute/TCPRoute linking the API Gateway and target service:
```yaml
kind: HTTPRoute
metadata:
  name: dashboard
  namespace: demoapp
spec:
  parentRefs:
    - name: apigw
      namespace: apigateway
  rules:
    - backendRefs:
      - kind: Service
        name: dashboard
        namespace: demoapp
        port: 9002
```

* The API Gateway will accept the route, and the backing ingress-gateway config entry will be updated to appropriately reference the target service. However, inspecting the envoy clusters endpoint for an individual API Gateway instance will show no known upstreams.

* Updating the newly-created namespace with the `cross-namespace-policy` automatically created by consul-k8s at cluster initialization will allow the API gateway's token to begin discovering service endpoints in other namespaces: `consul namespace update -name=apigateway -default-policy-name=cross-namespace-policy` (Note, this must be done via CLI or the HTTP API. The Consul UI seems to have a bug associating a new default policy to an existing namespace created in this manner. Perhaps due to the `external-source` meta applied to the namespace?)

### Expected behavior

API Gateway Controller should apply `cross-namespace-policy` to any created namespaces automatically, when ACLs and Namespaces are in use.

jm96441n:
When new namespace is created we should automatically apply the `cross-namespace-policy` policy to the newly created namespace to ensure cross namespace communication is allowed. 

Fixes #510 

### How I've tested this PR:
* Clone the files from the following gist: https://gist.github.com/jm96441n/7e005c82fa918b7b73c11abb617b4998
* Replace the `<YOUR ENTERPRISE KEY>` with a key for enterprise consul in the `deploy.sh` file
* From the api-gw repo run `make docker/dev` on the main branch (without these changes)
* Run the `deploy.sh` script
* After that finishes from the terminal run `curl 0:30602 -vv --header 'Host: nginx.green.daskt.nsbug.it'` and you should see a `503` error along with `no healthy upstream`
* In the consul UI if you select `Manage Namespaces` from the `Namespaces` drop down you will see the `green` namespace does not have the `cross-namespace-policy` attached to it
* In the api-gateway repo checkout this branch and run `make docker/dev` again to build the image with the changes from this branch
* Run the `reload.sh` script from the gist
* After that finishes from the terminal run `curl 0:30602 -vv --header 'Host: nginx.green.daskt.nsbug.it'` and you should see a response with a bunch of html (you may see a "connection reset by peer" error, if so give things a few seconds to reconcile and try again)
* In the consul UI if you select `Manage Namespaces` from the `Namespaces` drop down you will see the `green` namespace now has the `cross-namespace-policy` attached to it

Code Review
Can run the above steps

### Checklist:
- [ ] Tests added
- [X] CHANGELOG entry added 

### DIFF
--- a/internal/commands/server/command.go
+++ b/internal/commands/server/command.go
@@ -169,7 +169,7 @@ func (c *Command) Run(args []string) int {
 		consulCfg.Address = c.flagConsulAddress
 	}
 
-	partitionInfo := consul.NewPartitionInfo(os.Getenv("CONSUL_ENABLE_PARTITIONS") == "true", os.Getenv("CONSUL_PARTITION_NAME"))
+	partitionInfo := consul.NewPartitionInfo(os.Getenv("CONSUL_PARTITION"))
 
 	cfg.ConsulNamespaceConfig = k8s.ConsulNamespaceConfig{
 		ConsulDestinationNamespace:      c.flagConsulDestinationNamespace,
--- a/internal/commands/server/command.go
+++ b/internal/commands/server/command.go
@@ -60,7 +60,6 @@ type Command struct {
 	flagConsulDestinationNamespace string
 	flagMirrorK8SNamespaces        bool
 	flagMirrorK8SNamespacePrefix   string
-	flagPartitionInfo              consul.PartitionInfo
 
 	// Logging
 	flagLogLevel string
@@ -94,8 +93,6 @@ func (c *Command) init() {
 		c.flagSet.StringVar(&c.flagConsulDestinationNamespace, "consul-destination-namespace", "", "Consul namespace to register gateway services.")
 		c.flagSet.BoolVar(&c.flagMirrorK8SNamespaces, "mirroring-k8s", false, "Register Consul gateway services based on Kubernetes namespace.")
 		c.flagSet.StringVar(&c.flagMirrorK8SNamespacePrefix, "mirroring-k8s-prefix", "", "Namespace prefix for Consul services when mirroring Kubernetes namespaces.")
-		c.flagSet.BoolVar(&c.flagPartitionInfo.EnablePartitions, "enable-partitions", false, "[Enterprise Only] Enables Admin Partitions")
-		c.flagSet.StringVar(&c.flagPartitionInfo.PartitionName, "partition", "", "[Enterprise Only] Name of the Admin Partition")
 	}
 
 	{
@@ -172,11 +169,13 @@ func (c *Command) Run(args []string) int {
 		consulCfg.Address = c.flagConsulAddress
 	}
 
+	partitionInfo := consul.NewPartitionInfo(os.Getenv("CONSUL_ENABLE_PARTITIONS") == "true", os.Getenv("CONSUL_PARTITION_NAME"))
+
 	cfg.ConsulNamespaceConfig = k8s.ConsulNamespaceConfig{
 		ConsulDestinationNamespace:      c.flagConsulDestinationNamespace,
 		MirrorKubernetesNamespaces:      c.flagMirrorK8SNamespaces,
 		MirrorKubernetesNamespacePrefix: c.flagMirrorK8SNamespacePrefix,
-		PartitionInfo:                   c.flagPartitionInfo,
+		PartitionInfo:                   partitionInfo,
 	}
 
 	consulScheme, consulHTTPAddressOrCommand, port, err := parseConsulHTTPAddress()
--- a/internal/commands/server/command.go
+++ b/internal/commands/server/command.go
@@ -60,6 +60,7 @@ type Command struct {
 	flagConsulDestinationNamespace string
 	flagMirrorK8SNamespaces        bool
 	flagMirrorK8SNamespacePrefix   string
+	flagPartitionInfo              consul.PartitionInfo
 
 	// Logging
 	flagLogLevel string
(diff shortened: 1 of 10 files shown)


internal/commands/server/command.go
```diff
@@ -60,6 +62,7 @@ type Command struct {
 	flagConsulDestinationNamespace string
 	flagMirrorK8SNamespaces        bool
 	flagMirrorK8SNamespacePrefix   string
+	flagPartitionInfo              consul.PartitionInfo
```
andrewstucki, line 65: Just FYI, the addition of these flags means that previous versions of the controller will no longer work if we make a corresponding k8s change (since this is an unknown flag to them), so it's a breaking change. Personally just to make it so we don't really have to think about the support matrix for this, I'd prefer if we just passed these as environment variables.

internal/consul/namespaces.go
```diff
@@ -45,3 +66,76 @@ func EnsureNamespaceExists(client Client, ns string) (bool, error) {
...
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+// isPolicyExistsErr returns true if err is due to trying to call the
+// policy create API when the policy already exists.
+func isPolicyExistsErr(err error, policyName string) bool {
+	return strings.Contains(err.Error(), "Unexpected response code: 500") &&
```
andrewstucki, line 139: Can you add a quick comment here that this handles the case of ACLs not being enabled? It's always confusing that the only way to do this is to issue something against the policy endpoint and see if Consul blows up.

internal/consul/namespaces.go
```diff
@@ -29,8 +41,17 @@ func EnsureNamespaceExists(client Client, ns string) (bool, error) {
 		return false, nil
 	}
 
-	// If not, create it.
-	var aclConfig capi.NamespaceACLConfig
+	// If the namespace does not, create it with default cross-namespace-policy.
+	crossNamespacePolicy, err := getOrCreateCrossNamespacePolicy(client, partitionInfo)
```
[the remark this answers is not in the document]
jm96441n: yeah I don't think we want to patch existing namespaces for that reason you mentioned: they're not managed by us and might give broader permissions than the operator wants (it could be a bit of a footgun, but if the operator is manually creating namespaces I'd expect them to also be managing the permissions)

internal/consul/namespaces.go
jm96441n: done and done

internal/commands/server/command.go
jm96441n: move these to env variables

andrewstucki (approved):
We'll want to make a corresponding change in the Helm chart to set the environment variables. Feel free to link it back here.

jm96441n:
so @andrewstucki it looks like there might already be an env variable in the helm chart to handle that (should've checked their first) https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/templates/api-gateway-controller-deployment.yaml#L119 
going to update this PR to use that

PR #516, opened 2023-02-21
Update ACL token for e2e tests

jm96441n:
Create an [initial management token](https://developer.hashicorp.com/consul/docs/agent/config/config-files#acl_tokens_initial_management) for when we run e2e tests that has the correct permissions so tests against consul 1.15 pass.

### How I've tested this PR:
Running the e2e tests using the `make e2e consul_image="hashicorppreview/consul:1.15-dev"` command

Code review/running tests

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


mikemorris:
Looks like enterprise tests are failing consistently on `TestGatewayBasic/gateway_admission`, guessing something is still missing with namespace and/or partition scope for the updated ACL token.

jm96441n:
yeah, I'm looking into the enterprise tests, my guess is the ACL permissions need to be tweaked a bit

internal/k8s/builder/builder.go
```diff
@@ -31,7 +31,7 @@ func init() {
 }
 
 const (
-	defaultEnvoyImage           = "envoyproxy/envoy:v1.21-latest"
```
jm96441n, line 34: honestly I'm not a fan of making this change here, for some reason the envoy image specified in the e2e tests don't actually override this and I'm not 100% sure on what else I would need to change

internal/commands/server/k8s_e2e_test.go
```diff
@@ -2416,6 +2416,7 @@ func createGatewayClassWithParams(ctx context.Context, t *testing.T, resources *
 		Spec: apigwv1alpha1.GatewayClassConfigSpec{
 			ImageSpec: apigwv1alpha1.ImageSpec{
 				ConsulAPIGateway: e2e.DockerImage(ctx),
+				Envoy:            e2e.EnvoyImage,
```
jm96441n, line 2419: this sets the envoy image to use when running the gateway

andrewstucki (approved):
Overall changes LGTM, question about revving the actual default envoy image v. the one used solely in test.

internal/testing/e2e/service.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	envoyImage                = "envoyproxy/envoy:v1.21-latest"
+	EnvoyImage                = "envoyproxy/envoy:v1.24-latest"
```
andrewstucki, line 26: This seems fine -- this is pretty much just used in running our dummy services on the mesh, but since they're pulling xDS it's still a necessary change to get this running in test. That said, I'm assuming that changing the actual default value for envoy instances deployed as gateways will happen in a separate PR (that's the one you'll need to regen the golden files for)?

https://github.com/hashicorp/consul-api-gateway/blob/2a3f024662c0130f4cce76806bca69df9ed42efc/internal/k8s/builder/builder.go#L34

jm96441n:
Yup, going to be opening up a follow up PR to handle the non-test changes (I didn't want to modify non-test code as part of this PR which was just around fixing the tests themselves)

jm96441n:
for posterity this is the follow up PR: https://github.com/hashicorp/consul-api-gateway/pull/521

PR #537, opened 2023-03-21
Add partition identity to envoy deployment

andrewstucki:

Gateways deployed in non-default partitions weren't properly identifying themselves to xDS servers. This adds the required metadata into the envoy bootstrapping file to make sure that the envoy instance identifies itself properly.

### How I've tested this PR:

Unit tests + verified manually.

### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.changelog/537.txt
+++ b/.changelog/537.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix envoy deployments not properly identifying themselves when deployed to non-default partitions.
+```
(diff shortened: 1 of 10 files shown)


andrewstucki:
@mikemorris that may take a bit of time since I don't believe any of the e2e testing code has any sort of partition awareness. Do you mind if I try and knock that out as a follow-up? This is unit tested well enough with the actual bootstrap template checking that I'm confident we're specifying the needed parameter that was missing previously, and we did vet this on a real k8s deployment configured for a secondary partition.

PR #547, opened 2023-04-03
envoy: fix parse SPIFFE path parsing for non-default partitions

mikemorris:

Adds SPIFFE parsing logic to handle paths for non-default partitions in Consul Enterprise.

### How I've tested this PR:

- [x] Manual verification
  - Configuring a Kubernetes cluster in a non-default partition requires external Consul servers, which our existing e2e framework does not support. Followup work will be adding a test deploying Consul API Gateway in a non-default partition to the github.com/hashicorp/consul-k8s/ repo, similar to the basic [`TestPartitions_Connect`](https://github.com/hashicorp/consul-k8s/blob/main/acceptance/tests/partitions/partitions_connect_test.go#L30) example.

- Confirm that deploying a Consul API Gateway in a Kubernetes cluster mapped to a non-default partition in Consul Enterprise succeeds with this patch applied.
- Check whether existing tweaked logic is reasonable, or whether we may want to switch to RegEx parsing here.

### Checklist:
- [ ] ~~Tests added~~
- [ ] CHANGELOG entry added 

### DIFF
--- a/internal/envoy/middleware.go
+++ b/internal/envoy/middleware.go
@@ -103,20 +103,45 @@ func verifySPIFFE(ctx context.Context, logger hclog.Logger, store store.ReadStor
 }
 
 func parseURI(path string) (core.GatewayID, error) {
+	// TODO: would it be cleaner to switch this to RegEx parsing?
 	path = strings.TrimPrefix(path, "/")
-	tokens := strings.SplitN(path, "/", 6)
-	if len(tokens) != 6 {
-		return core.GatewayID{}, errors.New("invalid spiffe path")
+	tokens := strings.SplitN(path, "/", 8)
+
+	spiffePathErr := errors.New("invalid spiffe path")
+	var namespace, service string
+
+	if len(tokens) == 8 {
+		// Consul Enterprise in non-default partition
+		if tokens[0] != "ap" || tokens[2] != "ns" || tokens[4] != "dc" || tokens[6] != "svc" {
+			return core.GatewayID{}, spiffePathErr
+		}
+
+		// Gateway is always registered to a controller in the local Kubernetes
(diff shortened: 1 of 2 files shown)


internal/envoy/middleware.go
```diff
@@ -103,20 +103,45 @@ func verifySPIFFE(ctx context.Context, logger hclog.Logger, store store.ReadStor
...
 func parseURI(path string) (core.GatewayID, error) {
+	// TODO: would it be cleaner to switch this to RegEx parsing?
 	path = strings.TrimPrefix(path, "/")
-	tokens := strings.SplitN(path, "/", 6)
-	if len(tokens) != 6 {
-		return core.GatewayID{}, errors.New("invalid spiffe path")
-	}
-	if tokens[0] != "ns" || tokens[2] != "dc" || tokens[4] != "svc" {
-		return core.GatewayID{}, errors.New("invalid spiffe path")
+	tokens := strings.SplitN(path, "/", 8)
```
nathancoleman, line 108: Is there a useful link or comment that we could include explaining where the `8` comes from?

mikemorris:
```suggestion
        // https://pkg.go.dev/strings#SplitN
	tokens := strings.SplitN(path, "/", 8)
```

nathancoleman (approved):

internal/envoy/middleware.go
mikemorris: Not documented AFAIK, but the SPIFFE path can be found in Consul source at https://github.com/hashicorp/consul/blob/40854125a5cdabe3d19a4ef77486f8eeaf0b647c/agent/connect/uri_service.go#L39-L54

nathancoleman:
Tested manually w/ @andrewstucki @t-eckert and @mikemorris

PR #578, opened 2023-05-17
Use latest patch of Envoy 1.24 for all tests

nathancoleman:
Some tests were still using `envoy:v1.22-latest`. This updates us to `envoy:v1.24-latest` across the board.

### Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


nathancoleman:
It looks like we're intentionally using v1.22-latest with older versions of the Helm chart

PR #595, opened 2023-09-06
Fixes panic when nonexistent node is queried

missylbytes:
Adds a nil check for the call back from Consul.
### Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


.changelog/595.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Adds nil check to prevent panic on the call to Consul to get the list of services on a node
```
andrewstucki, line 2:
```suggestion
Fix a nil pointer panic when Consul returns a literal "null" when checking catalog nodes.
```

andrewstucki (approved):
Changed a bit of the wording, but should be good to go when we get tests to pass.

